PR #1, opened 2014-12-21
Adding some things I needed

amstan:
Please don't merge yet, this is just a request for comments. Some things are broken(see my notes).

I got it working perfectly with my other code though, it looks much neater than before too!

PR #2, opened 2015-01-08
Support for MIDI ports

mgeier:
This was brought up in #1, but I think it deserves a separate issue.

I initially only thought about audio ports, but of course it would also be nice to support JACK MIDI ports.

In this issue I'd like to collect some ideas about how to extend the current module to support MIDI ports in a meaningful, easy-to-use way.

### DIFF
--- a/examples/midi_chords.py
+++ b/examples/midi_chords.py
@@ -23,15 +23,14 @@ outport = client.midi_outports.register("output")
 def callback(frames, userdata):
     outport.clear_buffer()
     for offset, indata in inport.incoming_midi_events():
-        outdata = outport.reserve_midi_event(offset, len(indata))
-        # Note: potential errors are ignored! If len(outdata) == 0 -> error
-        outdata[:] = indata
+        # Note: This may raise an exception:
+        outport.write_midi_event(offset, indata)  # pass through
         if len(indata) == 3:
             status, pitch, vel = struct.unpack('3B', indata)
             if status >> 4 in (NOTEON, NOTEOFF):
                 for i in INTERVALS:
+                    # Note: This may raise an exception:
                     outport.write_midi_event(offset, (status, pitch + i, vel))
-                    # An exception will be raised if MIDI buffer is full
     return jack.CALL_AGAIN
 
 client.set_process_callback(callback)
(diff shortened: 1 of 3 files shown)


mgeier:
I added a commit with a possible implementation (14b43b90be1b99502352b120f8002c16ef8bcc36).

amstan:
> Is there any other "port type" used anywhere?

Interesting thought but:

     jack.JackError: 'midi_out': port registration failed

[gPortTypes only has 2 elements](https://github.com/jackaudio/jack2/blob/master/common/JackPortType.cpp). What if i want to use jack to transport my own weird data?? heh

> Something with more underscores?
That might be nice. midi_in_ports, and rename inports to be audio_in_ports (input perhaps?)

Otherwise it looks pretty good, feel free to merge it.

mgeier:
I don't understand your comment about the port registration failure, can you please clarify?

I don't know if one can actually invent a new port type as a user.
I've only seen audio and MIDI; and no trace of anything else.

If someone points out a use case I'm willing to consider it, but for now I think I'll stick with audio and MIDI and not exposing the actual "port type" and "port type ID" in the API.

About the list names ... I thought a bit about this and my current standpoint is this:

It seems that MIDI ports are kind of an add-on to JACK, and not really first-class citizens like audio ports.
Therefore, I think it makes sense to call the lists of audio ports `inports` and `outports` and then, as an add-on, call the MIDI ports `midi_inports` and `midi_outports`.
The same reasoning can be used for `Port` vs. `MidiPort` and `OwnPort` vs. `OwnMidiPort`.

IMHO, the names with one underscore look better then the ones with two underscores.

amstan:
>      jack.JackError: 'midi_out': port registration failed

> I don't understand your comment about the port registration failure, can you please clarify?
> 
> I don't know if one can actually invent a new port type as a user.
> I've only seen audio and MIDI; and no trace of anything else.

That was me trying to see if i can get another type of port that's not midi nor audio. Jack didn't let me and returned an error instead.

I'm ok with your other decisions. What I was saying about naming were mostly nits, feel free to ignore them.

Let's merge this pull request.

PR #3, opened 2015-01-17
Ringbuffer class

amstan:
Old discussion: https://github.com/amstan/jackclient-python/commit/9002ac903a96e5a7292be6cefef9987320394a4c

There's still some more stuff to do (from_buffer and documentation). But I followed some of the comments.

### DIFF
--- a/jack.py
+++ b/jack.py
@@ -187,6 +187,35 @@ void jack_set_error_function(void (*func)(const char*));
 void jack_set_info_function(void (*func)(const char*));
 void jack_free(void* ptr);
 
+/* ringbuffer.h */
+
+typedef struct {
+    char* buf;
+    size_t len;
+} jack_ringbuffer_data_t;
+typedef struct {
+    char* buf;
+    volatile size_t write_ptr;
+    volatile size_t read_ptr;
+    size_t size;
+    size_t size_mask;
+    int mlocked;
+} jack_ringbuffer_t;
+jack_ringbuffer_t* jack_ringbuffer_create(size_t sz);
+void jack_ringbuffer_free(jack_ringbuffer_t* rb);
+void jack_ringbuffer_get_read_vector(const jack_ringbuffer_t* rb, jack_ringbuffer_data_t* vec);
+void jack_ringbuffer_get_write_vector(const jack_ringbuffer_t* rb, jack_ringbuffer_data_t* vec);
+size_t jack_ringbuffer_read(jack_ringbuffer_t* rb, char* dest, size_t cnt);
+size_t jack_ringbuffer_peek(jack_ringbuffer_t* rb, char* dest, size_t cnt);
+void jack_ringbuffer_read_advance(jack_ringbuffer_t* rb, size_t cnt);
+size_t jack_ringbuffer_read_space(const jack_ringbuffer_t* rb);
+int jack_ringbuffer_mlock(jack_ringbuffer_t* rb);
+void jack_ringbuffer_reset(jack_ringbuffer_t* rb);
+void jack_ringbuffer_reset_size (jack_ringbuffer_t* rb, size_t sz);
+size_t jack_ringbuffer_write(jack_ringbuffer_t* rb, const unsigned char* src, size_t cnt);
+void jack_ringbuffer_write_advance(jack_ringbuffer_t* rb, size_t cnt);
+size_t jack_ringbuffer_write_space(const jack_ringbuffer_t* rb);
+
 /* transport.h */
 
 int  jack_transport_locate(jack_client_t* client, jack_nframes_t frame);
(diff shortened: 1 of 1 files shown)


jack.py
```diff
@@ -187,6 +187,35 @@
 void jack_set_info_function(void (*func)(const char*));
 void jack_free(void* ptr);
 
+/* ringbuffer.h */
```
mgeier: Please add an empty line here.

jack.py
```diff
@@ -187,6 +187,35 @@
...
+typedef struct {
+    char *buf;
```
mgeier: In the rest of the `cdef()` I changed the style to Stroustrup's C++ style (the asterisk is part of the type, not part of the variable).
Can you please also change this?

The style in the JACK headers is quite inconsistent, so I took the liberty of changing the style.

jack.py
```diff
@@ -187,6 +187,35 @@
...
+} jack_ringbuffer_data_t;
+typedef struct {
+    char *buf;
+    volatile size_t write_ptr;
+    volatile size_t read_ptr;
+    size_t size;
+    size_t size_mask;
+    int mlocked;
+} jack_ringbuffer_t ;
+
```
mgeier: This empty line is not necessary.

jack.py
```diff
@@ -187,6 +187,35 @@
...
+    size_t len;
+} jack_ringbuffer_data_t;
+typedef struct {
+    char *buf;
+    volatile size_t write_ptr;
+    volatile size_t read_ptr;
+    size_t size;
+    size_t size_mask;
+    int mlocked;
+} jack_ringbuffer_t ;
```
mgeier: Please remove the space before ;.

jack.py
```diff
@@ -187,6 +187,35 @@
...
+void jack_ringbuffer_free(jack_ringbuffer_t *rb);
+/* not implemented: jack_ringbuffer_get_read_vector, really fast uses only */
+/* not implemented: jack_ringbuffer_get_write_vector, really fast uses only */
+size_t jack_ringbuffer_read(jack_ringbuffer_t *rb, char *dest, size_t cnt);
+size_t jack_ringbuffer_peek(jack_ringbuffer_t *rb, char *dest, size_t cnt);
+void jack_ringbuffer_read_advance(jack_ringbuffer_t *rb, size_t cnt);
+size_t jack_ringbuffer_read_space(const jack_ringbuffer_t *rb);
+int jack_ringbuffer_mlock(jack_ringbuffer_t *rb);
+void jack_ringbuffer_reset(jack_ringbuffer_t *rb);
+/* not implemented: jack_ringbuffer_reset_size */
```
mgeier: Why not implement this?

This could be included in `reset()` by an optional size parameter: `def reset(self, size=None)`.

jack.py
```diff
@@ -1533,6 +1562,65 @@ def get_array(self):
...
+    return data
+
+def _returns_bytes(fun):
+    def wrapper(self, cnt):
+        ret_cnt, data = fun(self, cnt)
+        return bytes(_ffi.buffer(data, ret_cnt))
+    return wrapper
+
+class RingBuffer(object):
+    def __init__(self, size=256):
```
mgeier: I don't like the hard-coded default value, I'd prefer an obligatory argument.

Is there a strong reason why this should be 256?

jack.py
```diff
@@ -1533,6 +1562,65 @@ def get_array(self):
...
+        ptr = _lib.jack_ringbuffer_create(size + 1)
```
mgeier: I don't think your work-around is a good solution for the "problem".
I wouldn't change JACK's behavior but would instead make it clear in the documentation (see https://github.com/jackaudio/jack2/issues/94#issuecomment-70415209).

If users want a different behavior, they can add `+ 1` in their code.

jack.py
```diff
@@ -1533,6 +1562,65 @@ def get_array(self):
...
+        if not ptr:
+            raise JackError(
+                "Could not create ringbuffer of size {0}".format(size))
```
mgeier: I would remove the size information from the message.
I think `"Could not create ringbuffer"` is enough.

jack.py
```diff
@@ -1533,6 +1562,65 @@ def get_array(self):
...
+        self._ptr = _ffi.gc(ptr, _lib.jack_ringbuffer_free)
+
+    def _read(self, cnt):
+        data = _ffi.new("unsigned char[]", cnt)
+        ret_cnt = _lib.jack_ringbuffer_read(self._ptr, data, cnt)
+        return ret_cnt, data
+    read = _returns_bytes(_read)
+
+    def _peek(self, cnt):
+        data = _ffi.new("unsigned char[]", cnt)
```
x42: Allocating memory just to peek into a ringbuffer ?! Really?

Given that python still has a global lock, using a lock-free ringbuffer makes no sense in the first place.

jack.py
```diff
@@ -1533,6 +1562,65 @@ def get_array(self):
...
+        ret_cnt = _lib.jack_ringbuffer_peek(self._ptr, data, cnt)
+        return ret_cnt, data
+    peek = _returns_bytes(_peek)
+
+    def write(self, data):
+        data = _maybe_bytes(data)
```
mgeier: There should be no manual conversions inside of `write()`, CFFI already takes care of this.
It seems, though, that even in the current development version of CFFI, passing buffers to functions is not (yet?) supported. In this case, `ffi.from_buffer()` should be called outside of `write()`.

jack.py
x42: PS. I take that back, one-ended access (if the other end is in a RT safe code) can make some sense.

jack.py
```diff
@@ -1533,6 +1562,65 @@ def get_array(self):
...
+    @property
+    def write_space(self):
+        return _lib.jack_ringbuffer_write_space(self._ptr)
+
+    @property
+    def size(self):
+        return self._ptr.size - 1
+
+    def __repr__(self):
+        return "<jack.RingBuffer {0.read_space}/{0.size}>".format(self)
```
mgeier: Don't you want to add information about `write_space` (https://github.com/amstan/jackclient-python/commit/9002ac903a96e5a7292be6cefef9987320394a4c#commitcomment-9096324)?

I think read/write space is more interesting than size.

If you don't show `size`, this would probably make the examples you gave in https://github.com/jackaudio/jack2/issues/94 less awkward.

jack.py
mgeier: I asked if this will be supported in CFFI: https://bitbucket.org/cffi/cffi/issue/47/creating-a-cdata-from-a-buffer#comment-14921803

jack.py
mgeier: > Allocating memory just to peek into a ringbuffer ?! Really?

What else shall we do?
We have to pass something to `jack_ringbuffer_peek()` ...
And I guess we'll have to allocate memory for it at some point, I guess we cannot use a fixed-size array (or even a single `char`) on the stack in Python, right?
Shall we add an `out` argument to pass in pre-allocated memory?
Then we would also have to provide a function that allocates memory using `ffi.new()` and returns the cdata object.
Actually we wouldn't need an *additional* argument, since the length can be obtained from the cdata object.

Or is there something more reasonable that we can do?

jack.py
```diff
@@ -1533,6 +1562,65 @@ def get_array(self):
...
+        ret_cnt = _lib.jack_ringbuffer_peek(self._ptr, data, cnt)
+        return ret_cnt, data
+    peek = _returns_bytes(_peek)
+
+    def write(self, data):
+        data = _maybe_bytes(data)
+        return _lib.jack_ringbuffer_write(self._ptr, data, _ffi.sizeof(data))
+
+    def mlock(self):
+        return _lib.jack_ringbuffer_mlock(self._ptr)
```
mgeier: This should probably raise and exception on error, right?

I think you should use `_check()` here.

mgeier:
Good news everyone!
CFFI 0.9 was just released: https://pypi.python.org/pypi/cffi/0.9.0
Now `ffi.from_buffer()` should be officially available.

jack.py
amstan: Done

jack.py
amstan: Done

jack.py
amstan: Done

jack.py
amstan: Done

jack.py
amstan: Done.

Though what should we do about 
`@param sz the new size, that must be less than allocated size`
Have a check, just let it be?

jack.py
amstan: No strong reason, just wanted a quick way to define a buffer without specifying the size. Removing it.

jack.py
amstan: Done

jack.py
amstan: Done

amstan:
The new api is nice, but it still doesn't let me do things quickly without pain:

	>>> jack._ffi.from_buffer(b"123")
	Traceback (most recent call last):
	File "<stdin>", line 1, in <module>
	File "/usr/lib/python3.4/site-packages/cffi/api.py", line 281, in from_buffer
		return self._backend.from_buffer(self.BCharA, python_buffer)
	TypeError: from_buffer() cannot return the address of the raw string within a bytes or unicode or bytearray object

One needs this workaround:

	>>> jack._ffi.from_buffer(array.array("B",b"123"))
	<cdata 'char[]' buffer len 3 from 'array.array' object>

It's not a big deal, I'll use it anyway.

jack.py
mgeier: Since the JACK API doesn't care to return anything from this function, we shouldn't care either.

I think the description in the docs is enough, everything else is the user's problem.

mgeier:
I think that `from_buffer()` cannot be used with `bytes` isn't a problem.

`bytes` should be directly passed to the function, as well as (other) iterables of `int`.

I think it's best to wrap `from_buffer()` into a `try` block, then everything works.

See my comment at https://bitbucket.org/cffi/cffi/issue/47/creating-a-cdata-from-a-buffer#comment-15153192 and my implementation of `write_midi_event()`.

Your work-around shouldn't be necessary at all.

mgeier:
In addition, you should change the `cdef()` to `const unsigned char* src` and change `_ffi.sizeof()` to `len()`, see https://github.com/amstan/jackclient-python/commit/9002ac903a96e5a7292be6cefef9987320394a4c#commitcomment-9207146.

mgeier:
Thanks a lot for the updated PR, that's great!

I made a few comments, it would be nice if you could have a look at them.
Once those are resolved, I think we're ready to merge!

amstan:
> I think it's best to wrap from_buffer() into a try block, then everything works.
> In addition, you should change the cdef() to const unsigned char* src and change _ffi.sizeof() to len(), see amstan@9002ac9#commitcomment-9207146.

Done

Will send a new version shortly.

amstan:
Uploaded.

Sorry for taking so long by the way.

jack.py
amstan: I think having `space occupied` / `total space` is a little more in line to how people expect it to be. You can even do the division to get the percentage occupied.

If we change it then '/' might not make sense anymore. I guess we could use something like "<->". There will also be confusion on what's occupied and what's available.

The awkwardness is going to be there either way, write space at the beginning might still be bigger than the size argument.

mgeier: What about being more explicit?

    def __repr__(self):
        return ("<jack.RingBuffer with {0.read_space}/{0.write_space} bytes " +
                "read/write space>").format(self)

It would look like this:

    <jack.RingBuffer with 11/116 bytes read/write space>

BTW, I'm still not sure if it's a good idea to have this `__repr__()` at all ...

Given all the discussion, the awkwardness and possible confusion, we might be better off just using the default `__repr__()` and do `rb.read_space` and `rb.write_space` if we actually want to know the information?

amstan: On 13 May 2015 at 01:39, Matthias Geier <notifications@github.com> wrote:

> Given all the discussion, the awkwardness and possible confusion, we might
> be better off just using the default __repr__()
>
Alright, I removed it.

It might be nicer seeing the id than whatever I had since you can
differentiate between 2 buffers:
>>> jack.RingBuffer(100), jack.RingBuffer(100)
(<jack.RingBuffer object at 0x7fc6d3ed3240>, <jack.RingBuffer object at
0x7fc6d0a51ac8>)

Whereas before it was pretty boring:
(<jack.RingBuffer 0/128>, <jack.RingBuffer 0/128>)

Anything else? Or are we ready for merging?

Alexandru Stan
http://hypertriangle.com/~alex/

mgeier: OK, I think removing it is the best for now.

If we feel the need to have a custom `__repr__()` at some point in the future, we can discuss about adding it then.

I re-based and merged your commit, thanks!

mgeier:
I rebased the commit (7def7df26b2f2c2ed3d5dc4c3c2f76501a630e72) and merged it into master.

Thanks a lot @amstan!

PR #5, opened 2015-01-17
Allow process callback to be a c function

amstan:
Old discussions: https://github.com/amstan/jackclient-python/commit/6a9b75d6a3b64e6184ffe0a5f4ee7dbcabfe84b9

### DIFF
--- a/jack.py
+++ b/jack.py
@@ -739,12 +739,17 @@ class Client(object):
             is called.
 
         """
-        @self._callback("JackProcessCallback", error=STOP_CALLING)
-        def callback_wrapper(frames, _):
-            return callback(frames, userdata)
+        if isinstance(callback, _ffi.CData):
+            callback_wrapper = callback
+            arg = _ffi.cast("void *", userdata)
+        else:
+            @self._callback("JackProcessCallback", error=STOP_CALLING)
+            def callback_wrapper(frames, _):
+                return callback(frames, userdata)
+            arg = _ffi.NULL
 
         _check(_lib.jack_set_process_callback(
-            self._ptr, callback_wrapper, _ffi.NULL),
+            self._ptr, callback_wrapper, arg),
             "Error setting process callback")
 
     def set_freewheel_callback(self, callback, userdata=None):


jack.py
```diff
@@ -739,12 +739,21 @@ def set_process_callback(self, callback, userdata=None):
...
 
         """
-        @self._callback("JackProcessCallback", error=STOP_CALLING)
-        def callback_wrapper(frames, _):
-            return callback(frames, userdata)
+        if isinstance(callback, _ffi.CData):
+            if _ffi.typeof(callback) != _ffi.typeof("JackProcessCallback"):
+                raise ValueError("Expected %s, got %s" % \
+                                 (_ffi.typeof("JackProcessCallback"),
+                                  _ffi.typeof(callback)))
```
mgeier: This is too much error handling code for my taste.
You should check for both conditions in the same `if` expression and let the interpreter throw an error later if the type is wrong (in line 752).

amstan: I was trying to avoid a CData that's not a JackProcessCallback from being considered as a python callable.

mgeier: I know, but I think this isn't worth the effort for such an extremely uncommon use case.

The interpreter will throw an exception anyway (and it will not crash!) when it tries to call the thing.

I would only create a custom exception if its a very common error (which this is not) or when continuing would lead to nasty or subtle or hard-to-find errors later in the program (which this does not, because an exception is thrown by the interpreter and the process callback will not be called again).

amstan: Yep, you're right:

    TypeError: initializer for ctype 'int(*)(uint32_t, void *)' must be a pointer to same type, not cdata 'void(*)(char *)'

mgeier: That's great, getting rid of the check with `ffi.typeof()` is even better!
CFFI indeed does a lot of checking, so we should do it only where it's really necessary.

Just out of curiosity: will `ffi.typeof()` really call the C function? I thought the information about the signature is available in the cdata object ...

jack.py
```diff
@@ -739,12 +739,17 @@ def set_process_callback(self, callback, userdata=None):
             is called.
 
         """
-        @self._callback("JackProcessCallback", error=STOP_CALLING)
-        def callback_wrapper(frames, _):
-            return callback(frames, userdata)
+        if isinstance(callback, _ffi.CData):
+            callback_wrapper = callback
+            arg = _ffi.cast("void *", userdata)
```
mgeier: I would move the cast to `"void*"` out of here into the user code.
This way it would be more symmetrical: The user passes a `void*` and gets the same `void*` back within the callback function.

We should probably rename `arg` to `userdata_ptr` to make it a little clearer.
Then we can use

    callback_wrapper, userdata_ptr = callback, userdata

... here, and `userdata_ptr = _ffi.NULL` below.

What do you think?

At some point we might also consider renaming `callback_wrapper` to `callback_ptr` in the whole codebase, but I'm not yet sure if that's a good idea.

mgeier: I'm not quite sure, but I suspect the cast to `void*` isn't even necessary since CFFI seems to do the cast implicitly.

amstan:
Abandoning this, I changed my code to not need this and to just do cffi directly:
https://github.com/amstan/guitar/commit/d55fbf8c885a50b4786e526ea00d76cb785f3923#diff-7607b77999779e616acb8412424510deL59

PR #10, opened 2015-04-11
Add transport_query() method on Client

Vampouille:
Hello, 

I added transport_query() method in order to get current transport position and BPM.

I added several C types from /usr/include/jack/types.h
package : libjack-jackd2-dev 1.9.9.5+201306 (Linux Mint 17.1 64bit)
I also added a new attribute to Client class : 'position' to avoid malloc() in transport_query() method

### DIFF
--- a/jack.py
+++ b/jack.py
@@ -681,12 +681,9 @@ class Client(object):
                "Error locating JACK transport")
 
     def transport_query(self):
-        """return current transport position."""
-        # Test rolling transport status
-        if _lib.jack_transport_query(self._ptr, self._position) == 1:
-            return self._position
-        else:
-            return None
+        """return current transport state and position."""
+        transport_state = _lib.jack_transport_query(self._ptr, self._position)
+        return Transport(transport_state, self._position)
 
     def set_freewheel(self, onoff):
         """Start/Stop JACK's "freewheel" mode.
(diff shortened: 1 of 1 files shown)


mgeier:
Thanks a lot for this PR!

Can you please:

* put the C function declarations in the same order as they are in the original header files?
* remove the documentation from the C function declarations and move them to the appropriate Python function docstrings?
* remove whitespace from the C declarations to match the rest of the `cdef()` string?

jack.py
```diff
@@ -276,6 +355,7 @@ def __init__(self, name, use_exact_name=False, no_start_server=False,
 
         """
         status = _ffi.new("jack_status_t*")
+        self.position = _ffi.new("jack_position_t*")
```
mgeier: It's a good idea to allocate the memory during initialization, but the attribute shouldn't be public, you should probably call it `self._position` or similar.

It would probably be in better company a few lines further down, e.g. after `self._keepalive`?

jack.py
```diff
@@ -600,6 +680,14 @@ def transport_locate(self, frame):
         _check(_lib.jack_transport_locate(self._ptr, frame),
                "Error locating JACK transport")
 
+    def transport_query(self):
+        """return current transport position."""
+        # Test rolling transport status
+        if _lib.jack_transport_query(self._ptr, self.position) == 1:
+            return self.position
+        else:
+            return None
```
mgeier: Do you really want to return `None`?
Wouldn't raising an exception be more appropriate?
You could use the `_check()` function for that.

mgeier:
What about returning something more Pythonic by default (e.g. a `dict`), and providing an argument (e.g. `return_dict=False`) to return the CFFI struct for realtime usage?

The former could also do the consistency check `unique_1 == unique_2`.

jack.py
mgeier: I just saw that the return value isn't an error code but probably useful information in a `jack_transport_state_t`.
What about returning that, too?

jack.py
Vampouille: I change attribute name to '_position' and move initialization at the 
end of constructor.

jack.py
Vampouille: For my project, I just need to read transport position if transport is 
'rolling'. But you're right, we need return value of 
jack_transport_query() *and* content of position struct...

Vampouille:
Le 13/04/2015 20:22, Matthias Geier a écrit :
>
> What about returning something more Pythonic by default (e.g. a 
> |dict|), and providing an argument (e.g. |return_dict=False|) to 
> return the CFFI struct for realtime usage?
>
I just start python some days ago so I'm sure my code can be more 
Pythonic. You're help is welcome!

About dealing with CFFI struct for realtime usage :

My first idea was to hide CFFI struct with a class suitable for realtime 
usage that take CFFI struct as constructor argument. This class will 
have methods to access each fields (or group of fields) of the CFFI struct :

class Position():
     def __init__(self,cffi_data):
         self._data = cffi_data

     def get_bar_beat_tick(self):
         return (self._data.bar, self._data.beat, self._data.tick)

     def get_frame(self):
         return self._data.frame
     ...

constructor only store reference to the CFFI struct so it should be ok 
for realtime?
This class will avoid copy of all data ? (and it can be documented)

As you suggest, I will add a method to return a complete dict with all 
values in correct python type.

> The former could also do the consistency check |unique_1 == unique_2|.
>
Should I check consistency (|unique_1 == unique_2|) before returning 
position ?

Should I send you another pull request or can you check my fork ? I 
don't known if I can make modification on the current pull request.

Best,
Julien

jack.py
Vampouille: Le 13/04/2015 20:24, Matthias Geier a écrit :
>
> In jack.py 
> <https://github.com/spatialaudio/jackclient-python/pull/10#discussion_r28264312>:
>
> > @@ -600,6 +680,14 @@ def transport_locate(self, frame):
> >          _check(_lib.jack_transport_locate(self._ptr, frame),
> >                 "Error locating JACK transport")
> >
> > +    def transport_query(self):
> > +        """return current transport position."""
> > +        # Test rolling transport status
> > +        if _lib.jack_transport_query(self._ptr, self.position) == 1:
> > +            return self.position
> > +        else:
> > +            return None
>
> I just saw that the return value isn't an error code but probably 
> useful information in a |jack_transport_state_t|.
> What about returning that, too?
>
Yes I think better is to return a tuple ? :

	return_value = _lib.jack_transport_query(..., self.position)
         return (return_value, self.position)

  Maybe there is interesting values in position when transport is stopped

mgeier:
First of all, Python isn't really reliable for realtime stuff.
Many things may work quite well, but there isn't really a guarantee for anything.

I made these Python bindings mainly to be able to do the non-realtime stuff (connecting ports and so on).
I also implemented some of the realtime stuff, but mostly out of curiosity.

If you want a more reliable system, you should use C or C++.

That said, I still want to try to make the library as useful, reliable and efficient as is possible.

> My first idea was to hide CFFI struct with a class suitable for realtime 
usage that take CFFI struct as constructor argument.
> This class will 
have methods to access each fields (or group of fields) of the CFFI struct :

```Python
class Position():
     def __init__(self,cffi_data):
         self._data = cffi_data

     def get_bar_beat_tick(self):
         return (self._data.bar, self._data.beat, self._data.tick)

     def get_frame(self):
         return self._data.frame
     ...
```

Getter methods are un-Pythonic, you would use properties here:

```Python
    @property
     def bar_beat_tick(self):
         return self._data.bar, self._data.beat, self._data.tick
```

> constructor only store reference to the CFFI struct so it should be ok 
for realtime?
> This class will avoid copy of all data ? (and it can be documented)

It doesn't really make a difference if you create a class instance or a `dict`, because the class instance also holds a `dict` (or at least something similar).

So no, a class isn't better than a `dict`, because both will have to be allocated.

OTOH, Python uses some kind of memory pool for small objects, but I don't really know any details and I don't know if we should count on this.

> As you suggest, I will add a method to return a complete dict with all 
values in correct python type.

>> The former could also do the consistency check |unique_1 == unique_2|.
>
> Should I check consistency (|unique_1 == unique_2|) before returning position ?

I think this would be good, since that's what those strange fields are there for.

But if you make two different options - one for efficiency and one for Pythonicity - you should probably do this only on the latter.

> Should I send you another pull request or can you check my fork ?

You *can* make another pull request if you want, but normally you would just add more commits to your branch "add_transport_query".
When you push those commits to your repository, they will automagically be added to this PR.
But I guess you've found that out already ...

jack.py
mgeier: > Yes I think better is to return a tuple ?

Yes, I think returning a tuple would be good.

The question is if we can improve on the numbers in the enum `jack_transport_state_t`.
We can probably wrap this into a custom type that displays something like `jack.ROLLING` ...

mgeier:
BTW, you'll probably have to put some of the structs into a separate `cdef()` and use the `packed` argument on it, see https://cffi.readthedocs.org/.

jack.py
Vampouille: I make a new commit on my feature branch with class Transport. Now 
transport_query() method return a instance of class Transport. CFFI 
fields can be access with property or dict. dict is only created when 
transport_info() method is called.

About return value of _lib.jack_transport_query(), I put it in attribute 
'_state' of class Transport. Maybe I can just make '_state' attribute 
public : 'state' instead of creating a property (@property) ?

What do you think about my new commit ?

Julien

Vampouille: This is my test script :


#!/usr/bin/env python3

"""JACK client that prints all received MIDI events."""

import jack
import binascii

client = jack.Client("MIDI-Monitor")
port = client.midi_inports.register("input")


def callback(frames, userdata):
     for offset, data in port.incoming_midi_events():
         # TODO: use ringbuffer
         print("{0}: 0x{1}".format(client.last_frame_time + offset,
binascii.hexlify(data).decode()))
         print("Time : {}".format(client.frames_since_cycle_start))
         position = client.transport_query()
         print(position.transport_info)
     return jack.CALL_AGAIN

client.set_process_callback(callback)

with client:
     print("#" * 80)
     print("press Return to quit")
     print("#" * 80)
     input()

mgeier:
I think the `Transport` class you added in 7b6dfe9d80dc572f0484c3fbcdc78ed047fd2c23 adds far too much code for this little functionality.
There is a lot of boiler plate just to replicate the attributes which are already available on the CFFI struct object.

What about creating a `dict` by means of a `for` loop iterating over a list of attributes specified by you?

Or probably it's better to make a much simpler function (similar to what you had before) right now that only returns the CFFI struct (and the integer representing the transport state) and later, in a separate PR we try to make this more Pythonic?

Vampouille:
I just make a new commit without Transport class.

jack.py
```diff
@@ -92,6 +92,85 @@
     JackTransportStarting = 3,
     JackTransportNetStarting = 4,
 } jack_transport_state_t;
```
mgeier: Please add the declarations in the same order as in the original header files.
This avoids repeated declarations, like for example `jack_nframes_t`.

jack.py
```diff
@@ -92,6 +92,85 @@
...
+typedef uint64_t jack_unique_t;
+typedef uint64_t jack_time_t;
+typedef uint32_t jack_nframes_t;
+typedef enum {
+
+    JackPositionBBT = 0x10,     /**< Bar, Beat, Tick */
```
mgeier: Please remove the documentation from the `cdef()` string.
The documentation should be below in the docstrings (which I think it already is).

jack.py
```diff
@@ -1895,6 +2023,8 @@ def _hasflag(self, flag):
         return bool(self._code & flag)
```
mgeier: There should be only 2 blank lines.

Vampouille: I just update my rep with all modifications.

I finish a first version of my application. I'm able to play several 
loop (6 in tests) with jack configured with a ~40 ms latency. No xrun 
appears. With latency at 11ms there is some xruns around 10 per minute.

I have not notice any performance issue.

Julien

mgeier:
What about https://github.com/spatialaudio/jackclient-python/pull/10#issuecomment-92907063?

I think we should use the `packed` argument for the structs which have `PRE_PACKED_STRUCTURE` and `POST_PACKED_STRUCTURE`, right?

jack.py
```diff
@@ -92,6 +92,38 @@
     JackTransportStarting = 3,
     JackTransportNetStarting = 4,
 } jack_transport_state_t;
+typedef uint64_t jack_unique_t;
+typedef enum {
+    JackPositionBBT = 0x10,
+    JackPositionTimecode = 0x20,
+    JackBBTFrameOffset =      0x40,
+    JackAudioVideoRatio =     0x80,
```
mgeier: I would prefer a single space around the equals sign and a comma at the end of the last line.

jack.py
```diff
@@ -92,6 +92,38 @@
...
+    JackVideoFrameOffset =   0x100
+
```
mgeier: This newline is superfluous, right?

jack.py
```diff
@@ -92,6 +92,38 @@
...
+} jack_position_bits_t;
+struct _jack_position {
```
mgeier: I would prefer a single space character between the type and the name, I'm trying to establish this as a consistent style, see also #11.

Vampouille:
I add another cdef() with packed argument.

mgeier:
I squashed your commits into f8a21a83762db877b33952419c416f2a0264d2bc and a pushed a few follow-up commits.

Thanks a lot for this PR!

PR #8, opened 2015-05-17
Derive Status class from int

mgeier:


mgeier:
So I found a solution anyway ... but it's not as nice as I hoped ...

I'm not deriving from `int`, but most of the operators work with `int`s.

PR #13, opened 2015-05-21
Add support for Windows OS

Vampouille:
I add support for windows and tested with win7 32 and 64 bits. I have no OSX to test OSX version of jack.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -56,6 +56,9 @@ setuptools:
 Installation
 ------------
 
+Linux
++++++
+
 Once you have installed the above-mentioned dependencies, you can use pip_
 to download and install the latest release with a single command::
 
@@ -94,6 +97,25 @@ If you prefer, you can also replace the last command with::
 
 If you want to avoid installation altogether, you can simply copy ``jack.py``
 to your working directory (or to any directory in your Python path).
+
+Windows
++++++++
+
+First step is to install 'Jack Audio Conenction Kit' on Windows. On this page
+you should find a installer for your windows version :
+http://jackaudio.org/downloads/
+
+Then, for other python dependencies, you can install WinPython that already
+include Python, cffi and setuptools. Download a copy of WinPython here :
+http://winpython.sourceforge.net/
+
+Finnaly, you can install JACK-Client with pip. In WinPython installation
+directory, launch 'WinPython Command Prompt' and run::
+
+   pip install JACK-Client
+
+As mention previously, you can also download Jack-Client package from PyPI_ or
+get lastest source from Github_
 
 Usage
 -----
--- a/jack.py
+++ b/jack.py
@@ -27,6 +27,8 @@ __version__ = "0.2.0"
 
 import errno as _errno
 from cffi import FFI as _FFI
+import sys
+import platform
 
 _ffi = _FFI()
 _ffi.cdef("""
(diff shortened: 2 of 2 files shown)


mgeier:
It's great that it works on Windows, I didn't expect that!

Can you please write briefly how you installed JACK on Windows (or provide a link)?

Is there an installer or did you have to copy some files manually?

Where are the `*.dll` files installed to?

Or did you have to compile it yourself?

Vampouille:
Jack Audio Team provide a installer for both version of windows (32 and 64) on this page : http://jackaudio.org/downloads/. No need to install anything else or to compile, dlls are located in standard places. Most of the work was to find name of the dlls !

mgeier:
That's great, I didn't expect it to be so easy to install JACK on Windows ...

Regarding `dlopen()`, I think it would be good if you could do it similar to this:
https://github.com/bastibe/PySoundFile/blob/master/soundfile.py#L254

In other words:

* call `_lib = _ffi.dlopen("jack")` in a `try` block (without OS selection!)
* in the `except` block, check for the Windows stuff
  * I'm not quite sure which of the many (slightly different) ways to check for the OS to use ...
  * ... probably `platform.system()` and `platform.architecture()[0]`?
  * you can do the necessary import(s) right there, if it's not used anywhere else
* re-raise the exception for the non-Windows case

For OSX, this *should* work out of the box, but sometimes (for reasons unknown to me) it doesn't.
For now, I wouldn't make a special case for OSX (and certainly not disable it!).
Just because there is no special code handling OSX, doesn't mean it's not supported ...

If some OSX user out there knows a reliable solution for OSX, please speak up!

Vampouille:
ok, I will try to find somebody with a mac

Vampouille:
What about just testing some lib names and put os name only in comment :

libs = [ 'jack', 'libjack64', 'libjack']
#        linux,   win64,    win32

for lib in libs:
  try:
    dlopen(lib)
  except Error:
    continue
  break;

So If windows lib name change to jack.dll, no need to adjust jackclient-python. On the other hand, if windows contains a jack.dll, it will be loaded instead of jack's 'libjack.dll'. To address this issue we can test loaded lib with a simple call to jack_version() for example (Maybe this will add some unwanted computation)

What do you think about this ?

mgeier:
I wasn't aware of the potential confusion with another `jack.dll` file ...

The solution with the list doesn't really help there, does it?

I'd like to avoid calling into the library without good reason.
I guess it would be enough to just check with `hasattr()`, but any other `jack.dll` might as well have a function with the same name ... I guess this doesn't lead anywhere ...

Originally, I wanted to have a default case for proper OSes and then a backup-plan for Windows, but probably it's better to check for Windows first, and then do the rest in an `else` clause.

I think it's quite unlikely that they will change the DLL file name in the installer.

I think it's best to do it like you originally suggested, but turned around a bit.

You should check for Windows first (including the 32/64 bit stuff) and then do the "default" case in an `else` clause.
Let's not check any other OS unless we have special treatment for it.

In other words (untested!):

```Python
import platform as _platform

...

if _platform.system() == 'Windows':
    if _platform.architecture()[0] == '64bit':
        _lib = _ffi.dlopen("libjack64")
    else:
        _lib = _ffi.dlopen("libjack")
else:
    _lib = _ffi.dlopen("jack")
```

Would that work?

Vampouille:
Yes, I test it under win7 32 & 64 and linux

mgeier:
I squashed your commits (without docs changes) into 84160061ac9b5c06b27b960285c168a8810ffe23 and added a bit less documentation (bb5473a3c54a2acc4a45f0c8ae056bebef7a86d0) which is hopefully enough for Windows users.

I'm closing this, thanks for the commits and thanks for testing!

PR #14, opened 2015-05-21
Add set_timebase_callback()

Vampouille:
I add set_timebase_callback() method and set jack_transport_reposition() as TODO

### DIFF
--- a/jack.py
+++ b/jack.py
@@ -1248,6 +1248,104 @@ class Client(object):
             self._ptr, callback_wrapper, _ffi.NULL),
             "Error setting xrun callback")
 
+    def set_timebase_callback(self, callback, conditional=False,
+                              userdata=None):
+        """Register as timebase master for the JACK subsystem.
+
+        The timebase master registers a callback that updates extended
+        position information such as beats or timecode whenever
+        necessary. Without this extended information, there is no need
+        for this function.
+
+        There is never more than one master at a time. When a new client
+        takes over, the former callback is no longer called. Taking over
+        the timebase may be done conditionally, so it fails if there was
+        a master already.
+
+        The method may be called whether the client has been activated
+        or not.
+
+        Parameters
+        ----------
+        callback : callable
+            Realtime function that returns extended position
+            information. Its output affects all of the following process
+            cycle. This realtime function must not wait.
+
+            This function is called immediately after `process()` in the
+            same thread whenever the transport is rolling, or when any
+            client has requested a new position in the previous
+            cycle. The first cycle after `set_timebase_callback()` is
+            also treated as a new position, or the first cycle after
+            `jack_activate()` if the client had been inactive.
+
+            The timebase master may not use its `pos` argument to set
+            `pos.frame`. To change position, use
+            `jack_transport_reposition()` or
+            `jack_transport_locate()`. These functions are
+            realtime-safe, the `timebase_callback` can call them
+            directly.
+
+            It must have this signature::
+
+                callback(state:int,
+                         frames:int,
+                         pos:CFFI_struct,
+                         new_pos:bool, arg) -> None
+
+            * state : int
+                current transport state.
+            * frames : int
+                number of frames in current period.
+            * pos : Position CFFI structure
+                position structure for the next cycle; `pos.frame` will
+                be its frame number. If `new_pos` is `False`, this
+                structure contains extended position information from
+                the current cycle. If `True`, it contains whatever was
+                set by the requester. The `timebase_callback`'s task is
+                to update the extended information here. See `JACK
+                transport documentation`__ for the available fields.
+            * new_pos : bool
+                `True` (non-zero) for a newly requested `pos`, or for
+                the first cycle after the `timebase_callback` is
+                defined. `arg` the argument supplied by
+                `set_timebase_callback()`.
+            * arg :
+                the argument supplied by `set_timebase_callback()`.
+        conditional : bool
+            set to `True` for a conditional request. If set to `True`
+            this call will fail if there is already a timebase master
+        userdata :
+            an optional argument for the callback function.
+
+        Returns
+        -------
+        bool
+            `True` if new timebase master was set. `False` if
+            conditional request fail bacause another master is already
+            registered.
+
+__ http://jackaudio.org/files/docs/html/structjack__position__t.html
+
+        """
+        @self._callback("JackTimebaseCallback")
+        def callback_wrapper(state, frames, pos, new_pos, _):
+            return callback(state, frames, pos, new_pos, userdata)
+
+        return_code = _lib.jack_set_timebase_callback(self._ptr,
+                                                      conditional,
+                                                      callback_wrapper,
+                                                      _ffi.NULL)
+
+        if return_code == 0:
+            return True
+        elif conditional and return_code == -1:
+            return False
+        else:
+            raise JackError("{0} ({1})"
+                            .format("Error setting timebase callback",
+                                    return_code))
+
     def get_uuid_for_client_name(self, name):
         """Get the session ID for a client name.
 
--- a/jack.py
+++ b/jack.py
@@ -105,6 +105,7 @@ typedef enum {
 } jack_position_bits_t;
 /* _jack_position: see below in "packed" section */
 typedef struct _jack_position jack_position_t;
+typedef void (*JackTimebaseCallback)(jack_transport_state_t state, jack_nframes_t nframes, jack_position_t *pos, int new_pos, void *arg);
 /* deprecated: jack_transport_bits_t */
 /* deprecated: jack_transport_info_t */
 
@@ -236,11 +237,11 @@ size_t jack_ringbuffer_write_space(const jack_ringbuffer_t* rb);
 /* TODO: jack_release_timebase */
 /* TODO: jack_set_sync_callback */
 /* TODO: jack_set_sync_timeout */
-/* TODO: jack_set_timebase_callback */
+int jack_set_timebase_callback(jack_client_t* client, int conditional, JackTimebaseCallback timebase_callback, void* arg);
 int  jack_transport_locate(jack_client_t* client, jack_nframes_t frame);
 jack_transport_state_t jack_transport_query(const jack_client_t* client, jack_position_t* pos);
 jack_nframes_t jack_get_current_transport_frame(const jack_client_t* client);
-int  jack_transport_reposition(jack_client_t* client, const jack_position_t* pos);
+/* TODO: jack_transport_reposition */
 void jack_transport_start(jack_client_t* client);
 void jack_transport_stop(jack_client_t* client);
 /* deprecated: jack_get_transport_info */
--- a/jack.py
+++ b/jack.py
@@ -1331,7 +1331,8 @@ class Client(object):
                                               callback_wrapper,
                                               _ffi.NULL)
 
-        # Because of a bug in JACK2 (see TODO), we also check for -1:
+        # Because of a bug in JACK2 version <= 1.9.10, we also check 
+        # for -1:
         # See https://github.com/jackaudio/jack2/pull/123
         if conditional and err in (_errno.EBUSY, -1):
             return False
--- a/jack.py
+++ b/jack.py
@@ -1248,8 +1248,7 @@ class Client(object):
             self._ptr, callback_wrapper, _ffi.NULL),
             "Error setting xrun callback")
 
-    def set_timebase_callback(self, callback, conditional=False,
-                              userdata=None):
+    def set_timebase_callback(self, callback, conditional=False):
         """Register as timebase master for the JACK subsystem.
 
         The timebase master registers a callback that updates extended
@@ -1291,7 +1290,7 @@ class Client(object):
                 callback(state:int,
                          frames:int,
                          pos:CFFI_struct,
-                         new_pos:bool, arg) -> None
+                         new_pos:bool) -> None
 
             * state : int
                 current transport state.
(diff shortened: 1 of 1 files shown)


mgeier:
To be honest, I don't understand what exactly such a callback is supposed to do ...

Could you please give a simple example?

Probably something I can use for testing?

I would wait with this until #7 is done, because then you can drop all this useless `userdata` stuff.
I'll try to finish #7 this weekend.

Are you planning to implement `jack_transport_reposition()`, too?

If yes, I think it would make sense to do that first, otherwise it may be hard to test the callback, right?

BTW, what do you think about #12?

Vampouille:
This callback is used to set bpm and bbt (and bbt_offset) : position in bar beat ticks. Server only fill 'frame' field of position struct. Then callback have to set BPM and then compute bar, beat, ticks fields according to current bpm and frame count.
https://github.com/Vampouille/superboucle/blob/master/gui.py at the end of this file, you can see an implementation of timebase callback.
In this callback, I use QSpinBox to get BPM from gui, then I set BPM (beat per bar, ticks per beats, ...) and finally I compute bar, beat, ticks from bpm and frame count.

In the original version of jack-Client jack_transport_reposition() was *not* declared as 'TODO' but was not implemented, so I just turn declaration into 'TODO'.

In my application, I have feature to jump to a specified position in song in 'bar' (and change tempo). My first idea was to use jack_transport_reposition() function to change position or to change bpm. But this is not the way to do in jack because bpm will be overwritten on next process cycle by timebasemaster. For jumping in song I use transport_locate() function. I use timebase_callback to change bpm.

For me jack_transport_reposition() is useless.

Are you interesting in testing my app ? I'm looking for tester... and maybe my app can be useful for you.

jack.py
```diff
@@ -674,6 +675,38 @@ def transport_stop(self):
         """Stop JACK transport."""
         _lib.jack_transport_stop(self._ptr)
 
+    def set_timebase_callback(self,
+                              conditional, timebase_callback, userdata=None):
```
mgeier: Please change the order of `conditional` and `timebase_callback` and make former optional with `conditional=False` (I guess this is the right default value?).

Please also rename `timebase_callback` to `callback`, like in the other callback setters.

jack.py
```diff
@@ -674,6 +675,38 @@ def transport_stop(self):
...
+        There is never more than one master at a time. When a new client takes
+        over, the former timebase_callback is no longer called. Taking over the
+        timebase may be done conditionally, so it fails if there was a master
+        already.
+
+        The method may be called whether the client has been activated or not.
+
+        Parameters
+        ----------
+        conditional : int
```
mgeier: This should be `bool`.

jack.py
```diff
@@ -674,6 +675,38 @@ def transport_stop(self):
         """Stop JACK transport."""
         _lib.jack_transport_stop(self._ptr)
 
+    def set_timebase_callback(self,
```
mgeier: Please move the method to the other callback setters.
Probably as the last one after `set_xrun_callback()`?

jack.py
```diff
@@ -674,6 +675,38 @@ def transport_stop(self):
...
+        ----------
+        conditional : int
+            non-zero for a conditional request.
+        timebase_callback : callable
+            realtime function that returns position information.
+        arg :
+            an argument for the timebase_callback function.
+        """
+        @self._callback("JackTimebaseCallback")
+        def callback_wrapper(state, nframes, pos, new_pos, _):
```
mgeier: Please change `nframes` to `frames`.

jack.py
```diff
@@ -674,6 +675,38 @@ def transport_stop(self):
...
+        timebase may be done conditionally, so it fails if there was a master
+        already.
+
+        The method may be called whether the client has been activated or not.
+
+        Parameters
+        ----------
+        conditional : int
+            non-zero for a conditional request.
+        timebase_callback : callable
```
mgeier: Please add documentation about the function signature, like in the other callback setters.

You should also copy the relevant text from `types.h`.

jack.py
```diff
@@ -674,6 +675,38 @@ def transport_stop(self):
...
+        arg :
+            an argument for the timebase_callback function.
+        """
+        @self._callback("JackTimebaseCallback")
+        def callback_wrapper(state, nframes, pos, new_pos, _):
+            return timebase_callback(state, nframes, pos, new_pos, userdata)
+
+        _check(_lib.jack_set_timebase_callback(
+            self._ptr, conditional, callback_wrapper, _ffi.NULL),
+            "Error setting timebase callback")
```
mgeier: You should handle the `conditional=True` case without exception.

Probably we should return `True`/`False` if the callback was set or not?

jack.py
```diff
@@ -674,6 +675,38 @@ def transport_stop(self):
         """Stop JACK transport."""
         _lib.jack_transport_stop(self._ptr)
 
+    def set_timebase_callback(self,
+                              conditional, timebase_callback, userdata=None):
+        """Register as timebase master for the JACK subsystem.
+
+        The timebase master registers a callback that updates extended position
```
mgeier: Please limit the line length of docstrings to 72 characters, as of [PEP8](http://legacy.python.org/dev/peps/pep-0008/#maximum-line-length).

mgeier:
Thanks for the explanations.
I'm starting to understand how this is supposed to work, but I'm not sure about every detail yet ...

In the meantime, I closed #7 and #12, can you please update your PR accordingly?

I also added a few comments within the code.

> For me `jack_transport_reposition()` is useless.

OK, we can leave it as "TODO" for now.

jack.py
Vampouille: Yes. I have to check _lib.jack_set_timebase_callback() result. There is a EBUSY value, you use a "C #define" to define EEXIST but there is a python module that contains standard error code 'errno'. Can I use it ? or you don't want to add dependencies ?

Vampouille: It seems that jack lib never return EBUSY value. It returns -1 with a conditional request. I think there is a bug here : https://github.com/jackaudio/jack2/blob/master/common/JackClient.cpp#L859
It should test conditional param and return result variable.
I will create an issue but for jackclient-python I will make a dirty "if conditional and -1 ...." ?

Vampouille:
Can I rebase and force push my PR ?

jack.py
mgeier: Thanks for the hint about the `errno` module.
I wasn't aware of that module, but I saw that @amstan was also using it, so I cherry-picked his commit (05d6d21b232835a605744e54928cd4694038e505).

> or you don't want to add dependencies?

That's not a problem since that's in the standard library, so it isn't really a dependency.

You can just re-base your changes onto the current master, where the `errno` module is already imported (I just renamed it to `_errno` to not pollute the module namespace).

> I will make a dirty "if conditional and -1 ...." ?

There could still be a "normal" error and you wouldn't be able to distinguish between the "normal" error and conditional request with an already active timebase master, right?

I think we can just ignore the JACK bug.
If anybody actually has a problem with that, they should create a new issue.
And probably JACK1 does the right thing?

You will only use `conditional=False`, right?

Vampouille: Le 27/05/2015 14:08, Matthias Geier a écrit :
>
> In jack.py 
> <https://github.com/spatialaudio/jackclient-python/pull/14#discussion_r31126275>:
>
> > +        Parameters
> > +        ----------
> > +        conditional : int
> > +            non-zero for a conditional request.
> > +        timebase_callback : callable
> > +            realtime function that returns position information.
> > +        arg :
> > +            an argument for the timebase_callback function.
> > +        """
> > +        @self._callback("JackTimebaseCallback")
> > +        def callback_wrapper(state, nframes, pos, new_pos, _):
> > +            return timebase_callback(state, nframes, pos, new_pos, userdata)
> > +
> > +        _check(_lib.jack_set_timebase_callback(
> > +            self._ptr, conditional, callback_wrapper, _ffi.NULL),
> > +            "Error setting timebase callback")
>
> Thanks for the hint about the |errno| module.
> I wasn't aware of that module, but I saw that @amstan 
> <https://github.com/amstan> was also using it, so I cherry-picked his 
> commit (05d6d21 
> <https://github.com/spatialaudio/jackclient-python/commit/05d6d21b232835a605744e54928cd4694038e505>).
>
>     or you don't want to add dependencies?
>
> That's not a problem since that's in the standard library, so it isn't 
> really a dependency.
>
> You can just re-base your changes onto the current master, where the 
> |errno| module is already imported (I just renamed it to |_errno| to 
> not pollute the module namespace).
>
ok
>
>     I will make a dirty "if conditional and -1 ...." ?
>
> There could still be a "normal" error and you wouldn't be able to 
> distinguish between the "normal" error and conditional request with an 
> already active timebase master, right?
>
Yes
>
> I think we can just ignore the JACK bug.
> If anybody actually has a problem with that, they should create a new 
> issue.
>
I check hydrogen and qtractor code, no reference to EBUSY
>
> And probably JACK1 does the right thing?
>
> You will only use |conditional=False|, right?
>
In my app I set conditional to True in order to test : no problem. I 
would like my app to integrate with DAW like ardour or qtractor. But 
those application does not set bbt_offset field so I need to recalculate 
it with frame and bpm. No problem if bpm is constant over the song but 
if tempo change I need bbt_offset and daw set as master timebase. So in 
this case I will not register my app as timebase master and make a 
request with conditional = True.

mgeier:
It would be nice if `set_timebase_callback()` could also be used as a decorator, like the other callback setters.

This can be done by changing the signature to:

```Python
def set_timebase_callback(self, callback=None, conditional=False):
```

... and adding this in the very beginning of the function:

```Python
if callback is None:
    return lambda cb: self.set_timebase_callback(cb, conditional)
```

jack.py
mgeier: Actually, I was wrong, I think this should be `blocksize`, because that's what it is.

jack.py
mgeier: OK, if you need `conditional=True`, we'll have to work around the JACK2 bug.

But we still should check for `EBUSY`, because JACK1 might use it correctly and at some point JACK2 might get fixed.
Our code should just continue to work in this case.

jack.py
```diff
@@ -1247,6 +1248,98 @@ def callback_wrapper(_):
...
+            registered.
+
+__ http://jackaudio.org/files/docs/html/structjack__position__t.html
+
+        """
+        @self._callback("JackTimebaseCallback")
+        def callback_wrapper(state, frames, pos, new_pos, _):
+            return callback(state, frames, pos, new_pos)
+
+        return_code = _lib.jack_set_timebase_callback(self._ptr,
```
mgeier: In other places, the result variable is called `err`, you should also use this for consistency.

mgeier: ... and you should make the line break directly after (, then the rest would fit in 1 line.

jack.py
mgeier: I just tried it with JACK1, and it indeed returns `EBUSY` (+16).

jack.py
mgeier: I just realized that if you change the name according to my first comment, my second comment becomes obsolete, so just forget it!

jack.py
```diff
@@ -1247,6 +1248,98 @@ def callback_wrapper(_):
...
+                                                      _ffi.NULL)
+
+        if return_code == 0:
+            return True
+        elif conditional and return_code == -1:
+            return False
+        else:
+            raise JackError("{0} ({1})"
+                            .format("Error setting timebase callback",
+                                    return_code))
```
mgeier: I think the logic here is more complicated than necessary.
And you should use `_check()`.

What about this?

```Python
err = _lib.jack_set_timebase_callback(self._ptr, conditional,
                                      callback_wrapper, _ffi.NULL)

# Because of a bug in JACK2 (see TODO), we also check for -1:
if conditional and err in (_errno.EBUSY, -1):
    return False
_check(err, "Error setting timebase callback")
return True
```

You just have to add the URL to your bug report.

jack.py
```diff
@@ -1247,6 +1248,98 @@ def callback_wrapper(_):
...
+            `True` if new timebase master was set. `False` if
+            conditional request fail bacause another master is already
+            registered.
+
+__ http://jackaudio.org/files/docs/html/structjack__position__t.html
+
+        """
+        @self._callback("JackTimebaseCallback")
+        def callback_wrapper(state, frames, pos, new_pos, _):
+            return callback(state, frames, pos, new_pos)
```
mgeier: You should use `bool(new_pos)`.

jack.py
Vampouille: Nice, this remove reference to JackError

mgeier:
Thanks for the changes!

Did you already create an issue for JACK2?

I would like to add the issue URL as a comment in our code.

Vampouille:
Le 01/06/2015 17:09, Matthias Geier a écrit :
>
> Thanks for the changes!
>
> Did you already create an issue for JACK2?
>

https://github.com/jackaudio/jack2/issues/122

mgeier:
Wow, that was quick, it's been already merged into master!

You should close https://github.com/jackaudio/jack2/issues/122 and add https://github.com/jackaudio/jack2/pull/123 as a comment here in this PR.

Something like this:

```Python
        # Because of a bug in JACK2 version <= 1.9.10, we also check for -1.
        # See https://github.com/jackaudio/jack2/pull/123.
```

mgeier:
I squashed your changes into ce009fecdebfd00d3f7a88c910fd50bad93d7b88.

Thanks for the PR!

Vampouille:
I have a test file for `set_timebase_callback()` if you want

mgeier:
Sure, if it's instructive (and not too complicated), it would be nice to have it in the `examples/` folder.

Could you please make a new PR for it?

Vampouille:
yes

PR #16, opened 2015-05-25
Inheritable Method Callbacks

amstan:
This is an example what i meant by my comment at https://github.com/spatialaudio/jackclient-python/pull/4#issuecomment-77734369.

> More importantly, I think there is even a quite bad disadvantage, since all callbacks are called, even if they are empty functions. At least for the process callback, this adds unnecessary overhead.

The only function which could cause a performance penalty is process, and that one needs to be overridden anyway. While I could implement something that detects that nothing inherited most callbacks so don't set_*_callback on them, all the non-process callbacks happen very rarely so I don't think it's worth it.

> their code gets more explicit and more readable

Spamming set\_*\_callback in my \_\_init\_\_ is not my idea of readable when i have other stuff to worry about there too.

For the final merge I'm proposing everything under PythonicClient to the Client proper. It even has a way to disable the extra functionality by passing method_callbacks=False. I could work on the docstrings too a little so all set\_\*\_callback methods refer to the on_* methods for the callback signatures (and move the signatures there while I'm at it).

### DIFF
--- a/examples/more_pythonic_client.py
+++ b/examples/more_pythonic_client.py
@@ -1,0 +1,140 @@
+#!/usr/bin/env python3
+
+"""Create a JACK client that prints a lot of information with methods.
+
+This client is like chatty_client in functionality but also defines a class
+which has a lot of overridable methods (instead of using set_*_callback
+all the time).
+
+"""
+import jack
+
+class PythonicClient(jack.Client):
+    def __init__(self, name, method_callbacks=True, *args, **kwargs):
+        super().__init__(name, *args, **kwargs)
+        if method_callbacks:
+            self.set_method_callbacks()
+
+    def set_method_callbacks(self):
+        # NOTE: these are actually jack lib global, if one instanciates more
+        # than one clients, only the last one's methods will get these called
+        jack.set_error_function(self.error_function)
+        jack.set_info_function(self.info_function)
+
+        # per jack client instance callbacks
+        self.set_blocksize_callback(self.on_blocksize_change)
+        self.set_client_registration_callback(self.on_client_registration)
+        self.set_freewheel_callback(self.on_freewheel_change)
+        self.set_graph_order_callback(self.on_graph_order_change)
+        self.set_port_connect_callback(self.on_port_connect)
+        self.set_port_registration_callback(self.on_port_registration)
+        self.set_port_rename_callback(self.on_port_rename)
+        self.set_process_callback(self.process)
+        self.set_samplerate_callback(self.on_samplerate_change)
+        self.set_shutdown_callback(self.on_shutdown)
+        self.set_xrun_callback(self.on_xrun)
+
+    def error_function(self, message):
+        print("Jack Error:", message)
+
+    def info_function(self, message):
+        print("Jack Info:", message)
+
+    def on_blocksize_change(self, blocksize):
+        pass
+
+    def on_client_registration(self, name, register):
+        pass
+
+    def on_freewheel_change(self, starting):
+        pass
+
+    def on_graph_order_change(self):
+        pass
+
+    def on_port_connect(self, a, b, connect):
+        pass
+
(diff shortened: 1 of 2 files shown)


mgeier:
Thanks a lot for this PR!
I like some of the commits, but not all of them ...

aa5472be890ff0a1b3ab4cd4fd3510d8f8cc2b4e:

This is a really nice idea, I didn't like the return values very much myself.

But if we do this, we should go all the way and completely drop the return values from the callbacks.
If an exception occurs, CFFI will automatically return the value from the `error` argument, so we could unconditionally return the success value.
There would be no need for the module constants `SUCCESS` et al. anymore, they could become `_SUCCESS` or similar.

There is only one case where I'm unsure, namely the process callback.
I'm not sure if "don't call again" is really meant as an error or just as an equally reasonable alternative to "call again".

If a user wants to stop the callback without there being an exception already, she will have to raise an exception herself.
What exception shall we suggest our users to raise in this case?

A few exception classes that come to mind: `EOFError`, `RuntimeError`, `StopIteration` or simply `Exception`.
It doesn't really matter what they raise, but I think it would be good to provide a meaningful suggestion.

If you somewhat agree, it would be nice if you could make a separate PR for this, then we can discuss the details there.

91116296b5ecec8a64536374971a7d36e3338db2:

I like this idea!
Do we even need the separate property if we pass it to the callback?

479d37404c0da45b934860bda8ec5614d3ed7770:

This one I don't like.

I don't agree that this is Pythonic, on the contrary, I think it's much too Java-esque!

Forcing those poor callback functions onto a class for no good reason doesn't seem like a good idea to me.
Especially the example is a bad example (or rather good for what I'm trying to say), since not a single callback *method* uses `self`, so there is no reason for them to be methods in the first place.

I'm not saying that I want to forbid users to inherit from `jack.Client`, but if they do, they should register their callbacks explicitly.
I think it's also quite unlikely that they actually need *all* the callbacks.

> The only function which could cause a performance penalty is process, and that one needs to be overridden anyway.

That's not true (the first part is, but the second part isn't).
E.g. one could implement some kind of JACK connection manager, without needing the process callback.

While it's true that the only place where it actually matters performance-wise is the process callback, I still think that it's stylistically ugly to call all callback setters in the beginning (even if none is needed) and then call all those empty callbacks (even if none is needed).

> Spamming set_*_callback in my __init__ is not my idea of readable when i have other stuff to worry about there too.

That's a fair point, but I guess most users will only need a handful of callbacks [citation needed ...].
And those that need more can make a separate class like you showed in your commit, no need to mix the callback setup with other initializations ...
I just don't think it's the right thing to do within the library.

If we can find an example application where deriving from `jack.Client` actually makes a difference for the better, we can add it there.
But we shouldn't call it "Pythonic".

Just as a little side note: I don't like the names "`on_*()`" ... is there any "true" Python library (i.e. not something ported from another language) that uses this nomenclature?

> It even has a way to disable the extra functionality by passing `method_callbacks=False`.

I don't like this option.
If anything, it should be a separate class with a separate name, because the behavior is significantly different.

48b150e802c6ffee92dc136f7f4f4784539d656b:

I don't think that's an improvement at all, but of course that's highly subjective.
The scope of function arguments should be quite clear to the reader and I think using the same name in all functions actually makes them easier to read.

Is there any role model for your suggestion?

I guess we won't find anything like this in the standard library because the BDFL hates callbacks in general ...

mgeier:
What do you think about #24 as alternative to your first commit here?

What about my other comments?

mgeier:
I cherry-picked your second commit into #25, any comments or suggestions for improvements?

I've already merged #24.
I'm not planning to merge the rest of your suggestions for now (with the reasons given above), but I'm of course open for discussions!

mgeier:
@amstan Do you want to follow up on this?

If not, I'll close this at some point.

PR #19, opened 2015-06-04
add timebase master example

Vampouille:


mgeier:
That's great, thank you!

I think this helps a lot to understand how the timebase callback is supposed to work.

Just for reference, there is a similar example in the JACK sources:
https://github.com/jackaudio/example-clients/blob/master/intime.c

The code is still quite simple, so we can probably add a few more features ...?

What about passing the values (BPM etc.) as (optional) command line arguments?
See https://github.com/spatialaudio/jackclient-python/blob/master/examples/thru_client.py for how to do that.

What about user input during runtime?

Since we already use `input()`, we might as well use it for user input ...
The user could e.g. change the BPM by pressing `120<Return>` and quit the program by pressing `q<Return>`.
Or would this be too much?

Vampouille:
not too much, it would be great !

examples/timebase_master.py
```diff
@@ -0,0 +1,56 @@
...
+def frames2ticks(frame, ticks_per_beat, beats_per_minute, frame_rate):
+    ticks_per_second = (beats_per_minute * ticks_per_beat) / 60
+    return (ticks_per_second * frame) / frame_rate
+
+
+client = jack.Client("Timebase Master")
+
+
+@client.set_timebase_callback
+def timebase_callback(state, nframes, pos, new_pos):
```
mgeier: I think you should check `new_pos`, right?

You should also change `nframes` to `blocksize` (and you will actually need it in the implementation).

See https://github.com/jackaudio/example-clients/blob/master/intime.c.

examples/timebase_master.py
```diff
@@ -0,0 +1,56 @@
...
+import jack
+
+BPM = 130
+BAR_START_TICK = 0.0
+BEATS_PER_BAR = 4.0
+BEAT_TYPE = 4.0
+TICKS_PER_BEAT = 960.0
+
+
+def frames2ticks(frame, ticks_per_beat, beats_per_minute, frame_rate):
```
mgeier: Is this function really helpful/necessary/instructive?

Wouldn't it be simpler to just inline these two lines?

Vampouille:
I'm not agree with else part of `intime.c` (`if new_pos`). In my app, I never check `new_pos`, I always compute bar, beat, ticks from frame position, fps and bpm. I don't increment ticks based on blocksize. Maybe this part of code is faster .... but this "else" does not add any feature !

About new_pos parameter, I can print something when `new_pos` is `True` like 'Transport moved' or 'New Position'

About `frames2ticks()` function, yes I will 'inline' it

Vampouille:
bbt_offset field is missing maybe in future version. This is not implemented in ardour, hydrogen or qtractor but it is a must have in order to handle bpm modification in a song (from bar 90, bpm change from 120 to 135)

mgeier:
Thanks for the changes, the new features are great!

I still don't quite understand how the `new_pos` parameter should be handled.
The [documentation for  `JackTimebaseCallback`](http://jackaudio.org/files/docs/html/group__TransportControl.html#ga5861cf9b84a0a3c3f6ff0f193ca3ec2c) states:

> **pos**: address of the position structure for the next cycle; pos->frame will be its frame number. If new_pos is FALSE, this structure contains extended position information from the current cycle. If TRUE, it contains whatever was set by the requester. The timebase_callback's task is to update the extended information here.

So in the case of `new_pos` being `False`, shouldn't most of the fields already contain the right information?
Why are you re-assigning most of the values?
Is this necessary?

The docs say

> ... If TRUE, it contains whatever was set by the requester ...

They also speak about `jack_transport_reposition()` and `jack_transport_locate()` ... could a different client request, say, a different BPM value or different `beats_per_bar`?

Wouldn't you have to check for `pos.valid` in case of `new_pos` being `True`?

Sorry for the dumb questions, I'm just trying to understand ...

Vampouille:
At the beginning of callback pos struct have previous value. For example pos.tick was set to 43 and, after calculation, callback set pos.tick to 56.
If `new_pos` is `False` then we are sure that `pos` contains previous transport position. But we cannot use those information to compute new position because ticks are rounded. With `bbt_offset` and `bar`, `beat`, `tick` we can recompute position.

mgeier:
I see, thanks for the clarification:

I still don't understand this: what happens if a different client uses `jack_transport_reposition()`?
Wouldn't this be ignored in your current implementation?

Vampouille:
Yes, like other soft like Qtractor. My first try was to use `jack_transport_reposition()` to change bpm but other jack client overwrite this value, so I set timebase master.

mgeier:
OK, so if another client uses `jack_transport_locate()` it works and if it uses `jack_transport_reposition()` the request is ignored.

Is this correct?

If yes, I think it would be good to issue a notification message in the example client if a different client sets some data which is ignored.

Vampouille:
yes, it is correct.

In example, I can check `beat_per_minute` fields. If `new_pos` is `False` and user does not supply new value, `pos.bpm` should already have good value. I can also check ticks_per_beat, beat_per_bar, ... but I cannot check bbt fields beacause ticks are rounded.

mgeier:
OK. I think it would be good if you could ...

* only assign values where it's actually necessary
* use all provided values as long as it makes sense and it's not too complicated
* for all other values (which are currently ignored), print an information message

Vampouille:
I changed example, now, bpm and others static fields : bar_start_tick, beats_per_bar, .... are overwrite only when new_pos is True. If new_pos is False I test if those fields are modified.

I test with ardour and qtractor, it seems that when you jump in song those application or jack reset pos struct to a fresh new memory emplacement with strange value : 

    'bar start tick' modified by another app 0.0 -> 1e-323
    'beats per bar' modified by another app 4.0 -> 0.0
    'beat type' modified by another app 4.0 -> 5.324934164434305e-44
    'ticks per beat' modified by another app 960.0 -> 2.81713227e-316
    'beats per minute' modified by another app or user 130 -> 0.0

`state` value is still 1 (playing) `pos.frame` is set to zero only by qtractor which produce a division by zero. Ardour set ticks per beat very low so it is rounded to zero too.

With previous version, I still have division by zero with qtractor just with a mouse click on tempo spin box to set focus on it.

I find nothing on internet about frame rate set to zero.

mgeier:
Thanks for your changes.

But now it doesn't seem to work anymore ...

> ... are overwrite only when `new_pos` is `True`.

But aren't you supposed to check the values in this case?

I think you should check `pos.valid`, right?

It's very strange that the `frame_rate` is zero, especially since the [docs](http://jackaudio.org/files/docs/html/structjack__position__t.html) state: "these cannot be set from clients; the server sets them".

Vampouille:
I reorganize callback, now I check about modified fields only when pos.valid have BBT flag set. I always set `pos` fields. 
I think the good question is how to test this ? I think we need an external software like qtractor or ardour that use transport_reposition() function....

examples/timebase_master.py
```diff
@@ -0,0 +1,110 @@
...
+    if new_pos:
+        print("Position changed")
+
+    if not pos.frame_rate:
+        print("Frame position : %s" % pos.frame)
+        print("Frame rate : %s" % pos.frame_rate)
+        print("State : %s" % state)
+        print("Valid : %s" % pos.valid)
+
+    if pos.valid & 0x10:
```
mgeier: I think you should check this only if `new_pos` is true.
If not, I guess we should assume all fields except the first 3 contain whatever we assigned before, right?

examples/timebase_master.py
```diff
@@ -0,0 +1,110 @@
...
+        check_modified_fields(pos)
```
mgeier: If any of the other "valid" flags are set, you should print a message, because we are ignoring this case.

examples/timebase_master.py
```diff
@@ -0,0 +1,110 @@
...
+    pos.bar = bar + 1
+    pos.beat = beat + 1
+
+    # Set valids fields
+    pos.valid = 0x10  # BBT Fields
+
+    return None
+
+
+def check_modified_fields(pos):
```
mgeier: Currently, you are ignoring all the BBT fields which are changed by some other client.
Don't you want to use those values for your calculations?

mgeier:
> how to test this ?

That's indeed a good question!

> I think we need an external software like qtractor or ardour that use transport_reposition() function....

But do they respect if another client wants to be timebase master?
I guess they are typically timebase masters themselves, right?

Would #22 help?

mgeier:
I merged #22 into master.

mgeier:
@Vampouille If you are still interested in this, I've just added an implementation of the "showtime" client, see #20.

Vampouille:
For me, the purpose of this PR is to show an example of computation of `bar`, `beat` and `ticks` from `pos.frame`. Interaction with other tools are maybe out of scope of this example. "Showtime" client may help in understanding interaction. So maybe we can add some comments in this example to explain limitation.

mgeier:
> Interaction with other tools are maybe out of scope of this example.

I think we shouldn't ignore other tools. Especially with the handling of `new_pos`, see my comment above: https://github.com/spatialaudio/jackclient-python/pull/19#issuecomment-110320257.
I think `new_pos` can only be understood if there is another client interacting, right?
Or probably it can be the same client making those changes ... but then we would need an extended user interface that allows such changes in addition to changing the BPM.

What about my other questions/comments?

It would be nice to get this example into `master`, but I think a few things should be changed before.

Vampouille:
Thanks @mgeier !

Issue #30, opened 2015-10-12
"timeout waiting for client" from jack daemon when using libjack calls from within port_registration_callback


mk-fg:
Recently had an issue, where: 

* port_registration_callback (set via `set_port_registration_callback()` before `activate()`) function was running a bunch of libjack (through this module, ofc, [code link](https://github.com/AccelerateNetworks/PagingServer/blob/2652566/paging.py#L483)) calls.
* same-name port was registered twice during app lifetime.

And that caused client (current git master - c5982a39) to reliably crash with:
```
cannot read result for request type 10 from server (Connection reset by peer)
cannot send event response to engine (Broken pipe)
cannot continue execution of the processing graph (Bad file descriptor)
jack_client_thread: graph error - exiting from JACK
```

And JACK1 (0.124.1, Debian Jessie) logging something like:
```
Oct 11 06:43:42 Paging-Test6 jackd1-paging[801]: timeout waiting for client paging.812 to handle a port registered event
Oct 11 06:43:42 Paging-Test6 jackd1-paging[801]: cannot send port registration notification to paging.812 (No such file or directory)
```

This seem to be 100% reproducible in these kinda-specific conditions - i.e. when it's same port name being registered for the second time during client lifetime - no other calls to that callback seem to trigger it.
Can be easily reproduced with that particular app (before it was fixed not to call stuff from callbacks) by running alsa_out, closing it, then running it again - client drops with above symptoms.

Also, all `connect()` / `disconnect()` calls from within that callback seem to raise JackError, while actually creating/removing connections, for any ports.

So I wonder, is it expected (e.g. "calling stuff from callbacks is obviously bad!") or sensible behavior?

And if it is, maybe it should be documented under set_port_registration_callback?
I.e. something like "calling other libjack functions here may lead to unexpected behavior like deadlocks or client timeouts" there.
If so, I can probably create a PR, but given relative lack of JACK/client knowledge, I suspect I might not be best at describing what exactly should be avoided here.

mgeier:
Thanks for the report.
I'm not a JACK expert myself, I just took the documentation from the comments in the JACK header files and moved them to the Python docstrings, I hardly added anything myself.

Basically, I only know what's in the documentation. For the case of [set_port_registration_callback()](http://jackclient-python.readthedocs.org/#jack.Client.set_port_registration_callback), it only says that "... the supplied function does not need to be suitable for real-time execution".
It doesn't say if it is allowed to call other JACK API functions from there.
[...]

PR #31, opened 2015-10-19
DOCS: add warnings about interacting with jack daemon from callbacks

mk-fg:
As mentioned in #30 and [on the jack-devel mailing list](http://thread.gmane.org/gmane.comp.audio.jackit/29271/focus=29273):

    You cannot make any calls that talk to the server from within a callback from the server.

This PR adds a "Warning:" to every non-RT (where such calls seem to be expected) callback function description and same info to the README section that mentions callbacks.

### DIFF
--- a/jack.py
+++ b/jack.py
@@ -846,6 +846,9 @@ class Client(object):
                statement*, which takes care of activating, deactivating
                and closing the client automatically.
 
+            .. note:: Same as with most callbacks, no functions that
+               interact with the JACK daemon should be used here.
+
         """
         @self._callback("JackInfoShutdownCallback")
         def callback_wrapper(code, reason, _):
@@ -944,6 +947,9 @@ class Client(object):
             The argument `starting` is ``True`` if we start to
             freewheel, ``False`` otherwise.
 
+            .. note:: Same as with most callbacks, no functions that
+               interact with the JACK daemon should be used here.
+
         See Also
         --------
         set_freewheel
@@ -991,6 +997,9 @@ class Client(object):
                previously referenced, and perform other operations that
                are not realtime safe.
 
+            .. note:: Same as with most callbacks, no functions that
+               interact with the JACK daemon should be used here.
+
         See Also
         --------
         :attr:`blocksize`
@@ -1032,6 +1041,9 @@ class Client(object):
             The argument `samplerate` is the new engine sample rate.
             The `callback` is supposed to raise :class:`CallbackExit` on
             error.
+
+            .. note:: Same as with most callbacks, no functions that
+               interact with the JACK daemon should be used here.
 
         See Also
         --------
@@ -1075,6 +1087,9 @@ class Client(object):
             argument is ``True`` if the client is being registered and
             ``False`` if the client is being unregistered.
 
+            .. note:: Same as with most callbacks, no functions that
+               interact with the JACK daemon should be used here.
+
         """
         @self._callback("JackClientRegistrationCallback")
         def callback_wrapper(name, register, _):
@@ -1110,6 +1125,9 @@ class Client(object):
             :class:`OwnPort` or :class:`OwnMidiPort` object, the second
             argument is ``True`` if the port is being registered,
             ``False`` if the port is being unregistered.
+
+            .. note:: Same as with most callbacks, no functions that
+               interact with the JACK daemon should be used here.
 
         See Also
         --------
@@ -1151,6 +1169,9 @@ class Client(object):
             objects of the ports which are connected or disconnected.
             The third argument is ``True`` if the ports were connected
             and ``False`` if the ports were disconnected.
+
+            .. note:: Same as with most callbacks, no functions that
+               interact with the JACK daemon should be used here.
 
         See Also
         --------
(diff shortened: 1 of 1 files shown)


mk-fg:
I did test same (only slightly modified) example as posted on the ML with client_registration_callback and port_connect_callback btw, all seem to time-out in exactly same fashion.

mgeier:
Thanks for the PR and thanks for testing this, that's really good to know.

Did you also try it with the blocksize callback?
This seems to be a somewhat special case ...

I think it's not good to add this stuff to the README, it only complicates matters.
In fact, I just removed some details from the README: 092623ac9f6e38f6382b4a8f6b503bb3956c0e53.

I find a "warning" for this a bit heavy, a "note" would surely be enough.
Probably it would be even enough to write it as a separate paragraph without any special decoration.

Actually, I'm thinking about removing the "note" markup from this sentence from the `set_*_callback()` methods:
*"This function cannot be called while the client is activated (after activate() has been called)."*
That many "note" blocks may become a bit annoying, what do you think?

jack.py
```diff
@@ -842,6 +842,8 @@ def set_shutdown_callback(self, callback):
                shutdown callback and has to be called outside of the
                callback context.
 
+            .. warning:: Same as with most jack callbacks, no functions
+               that iteract with jack daemon should be used here.
```
mgeier: Please add an empty line before the closing `"""`, as this seems to be the NumPy style.

jack.py
```diff
@@ -940,6 +942,9 @@ def set_freewheel_callback(self, callback):
             The argument `starting` is ``True`` if we start to
             freewheel, ``False`` otherwise.
 
+            .. warning:: Same as with most jack callbacks, no functions
+               that iteract with jack daemon should be used here.
```
mgeier: iteract -> interact (or communicate)?

jack.py
```diff
@@ -987,6 +992,9 @@ def set_blocksize_callback(self, callback):
                previously referenced, and perform other operations that
                are not realtime safe.
 
+            .. warning:: Same as with most jack callbacks, no functions
+               that iteract with jack daemon should be used here.
```
mgeier: jack daemon -> the JACK daemon

mk-fg:
> I think it's not good to add this stuff to the README, it only complicates matters.

Yeah, I agree, this probably belongs to callback docs.

> I find a "warning" for this a bit heavy, a "note" would surely be enough.

Don't really have opinion either way - I think it can be a warning, note, or part of a spec/description, depending on how you phrase it.

> That many "note" blocks may become a bit annoying, what do you think?

I think separate "note" blocks make it way easier to process these tips, i.e. you don't glaze over them as one text chunk, but rather take mental notes of each one individually.

Plus, it's a good markup for when you only want to know what the method is about (i.e. description) and whether you want it (or e.g. some other method), and don't care about caveats (unless it's the one you want), so I think it's actually better to move stuff like `All “notification events” are received ...` into such notes, and keep generic descriptions in the same way you'd have first line of a git commit.

Note that this might not be well-informed or even thought-out opinion.

> iteract -> interact (or communicate)?

Yeah, definitely a typo, copy-pasted all over the place ;(

Noted other comments as well.

------

Given that you want to shuffle stuff around anyway, maybe it'd be easier for you to just put these wherever you like (as I suspect reviewing this PR already took more of your time than just copy-pasting that line properly yourself)?

If not, can definitely do it, would you like it to be a warning, note or part of a description though?

mgeier:
Thanks for your comments!

> reviewing this PR already took more of your time than just copy-pasting that line properly yourself

It's not about me telling you what's *proper* (because I don't know that yet), I'm more interested in some discussion and your statements already helped me a lot.

It sounds reasonable to keep the notes which are already there, and I now agree with you that the new repeated "warnings" should also get their own boxes.

But I still think actual "warnings" are too strong, I think I would prefer "notes".

There may be cases where two notes come next to each other, but I guess it's no problem to have multiple notes in a row, right?

If you agree, could you please rebase and amend your PR accordingly?

mk-fg:
> There may be cases where two notes come next to each other, but I guess it's no problem to have multiple notes in a row, right?

I think it's better than two paragraphs, as it's easier to skim through these and there's no ambiguity about their relation to each other.

> If you agree, could you please rebase and amend your PR accordingly?

Yeah, sure, will do.

mk-fg:
Rebased, changed the message as suggested.

> Did you also try it with the blocksize callback?
> This seems to be a somewhat special case ...

Just tried with mod to the same script - https://gist.github.com/mk-fg/392ad642a33956950ff6 - disconnects in similar fashion on the first call into that callback, apparently.

mgeier:
Cool, thanks for the changes and for testing!

Merging ...

If you have further suggestions for improvements, don't hesitate to speak up!

mk-fg:
Thanks!
Will do, if I'll ever get to use JACK in a python project again.

mgeier:
Great!

If you need a more cross-platform audio solution for Python (using PortAudio), you should check out https://github.com/spatialaudio/python-sounddevice!

PR #33, opened 2016-05-09
Implement jack_get_current_transport_frame

mgeier:
Probably as a read-only property named `transport_frame`?
Or probably combine it with `transport_locate()` to a read/write property?


mgeier:
I've implemented this in f1a9f611b512d1b202c51f074cdff39351e61a31. If there are no objections, I'll merge this soon.

PR #34, opened 2016-07-04
WIP: Playback example with NumPy

LibrEars:
Is it possible to add an playback example with numpy. This pull request is how I think it is somehow done.

For my application i need low-latency multichannel output/recording (reference: https://github.com/bastibe/PySoundCard/issues/59). As I understand it, jack should be a good choice for that?

### DIFF
--- a/examples/allOutputs.py
+++ b/examples/allOutputs.py
@@ -1,0 +1,63 @@
+#!/usr/bin/env python3
+
+"""
+Create a JACK client that plays a soundfile to all outputs (loudspeakers and headphones).
+"""
+
+import sys
+import signal
+import os
+import jack
+import threading
+import numpy as np
+import soundfile as sf
+
+data, fs = sf.read('tschirp.wav')
+
+
+if sys.version_info < (3, 0):
+    signal.signal(signal.SIGINT, signal.SIG_DFL)
+    # See other examples
+else:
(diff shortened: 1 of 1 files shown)


mgeier:
If you want a simple way to do that, you should try `play()`, `rec()` and/or `playrec()` from the [sounddevice](http://python-sounddevice.rtfd.io/) module.

One of the [example programs](http://python-sounddevice.readthedocs.io/en/latest/examples.html) shows how to load a sound file and play it back.

The `sounddevice` module uses PortAudio, so you can still use JACK if you want. However, you can also use all the other host APIs, which will make your code more portable.

mgeier:
@LibrEars I've just added a file playback example (using NumPy) in #38. What do you think about it?
Any suggestions for improvements?

LibrEars:
Looks nice. Thanks a lot. I try it out and report

mgeier:
@LibrEars I changed the implementation of #38, making it significantly simpler (and a little less efficient).

mgeier:
@LibrEars Do you want to further pursue this at some point?

If not, I'd like to reject this PR.

LibrEars:
@mgeier sorry for not replying. You can reject the PR. But thank you again for the example! I still want to try it out, but couldn't taken the time I for it. -.-

Issue #35, opened 2016-08-29
SegFault on port_registration_callback for existing clients on quit


MaurizioB:
There is a strange behaviour for clients existing *before* a new jack.Client is created, once the port registration callback is set.
I made a test using python 2.7, ardour and jack_mixer.
1) launch the application; it has to have both audio and midi inputs and outputs. For example, this doesn't happen when launching jack_mixer without any channel set.
2) create a client with a port registration callback; these commands will suffice:

    client = jack.Client('xyz', no_start_server=True)
    def port_cb(*args):
        print 'port_cb: {}'.format(args)
    client.set_port_registration_callback(port_cb)
    client.activate()
3) close the former application

This generates a segfault on my machine (unfortunally I'm not able to make tests on other setups). The strange thing is that this does *not* happen when the application is started after the jack.Client instance is created and activated.
This problem is not reproducible when using a client created with jackclient-python (I don't exactly know why, maybe they exit in a different way).

mgeier:
Thanks for this nicely reproducible bug report!

The problem seems to be that the unregistered port doesn't exist anymore at the point when the callback is invoked.

Before the user-specified callback is called, the function `jack_port_by_id()` is called, which returns `NULL` if the port doesn't exist anymore. This is not checked and causes the segfault later.

Apart from the port registration callback, this might also affect the port connect callback. Probably even the port rename callback, if the port is destroyed right after renaming it.

The question is now: what should happen instead of the segmentation fault?

* Call the callback function with `None` instead of a `Port` object?
* Silently ignore the event and don't call the callback at all?
* Raise an exception?
* Something else?

MaurizioB:
You're welcome! By the way, thank you for this module! :)
Well, since I'm not a programmer, I'm not really sure what should be the better solution here, so I'm giving you just some ideas as a user who has fun with python.

I think that ignoring the event or raising an exception can be confusing: if I set a callback for that, I'd expect it to behave everytime in the same way, whether the other client's ports have been created before or after. Calling with None can be confusing too, because I wouldn't know why some callbacks would have the Port object and others wouldn't. Consistency should be important.

I just made a small test setting the connection callback too: I started the other client, I activated the jack.Client, I connected/disconnected every port in the other client, and then I close it. No SegFault.
[...]

PR #37, opened 2016-10-09
Updated port callbacks notes

MaurizioB:
I added some information about the callback behavior as discussed on #35, suggesting a workaround to avoid the need for setting `only_available`.


mgeier:
@MaurizioB Thanks a lot!

PR #45, opened 2018-01-18
Implement Port.alias() property.

jofemodo:
Uses "int jack_port_get_aliases(const jack_port_t * port, char *const aliases[2])" from the jack C library.


mgeier:
Thanks for this PR!

What about adding `set_alias()` and `unset_alias()`?
This way we could easily test if/how the `alias` property works.

There seem to be up to two aliases per port, what about returning a `list` of `str` instead of a single `str`?

src/jack.py
```diff
@@ -1440,6 +1440,24 @@ def shortname(self):
...
+        The alias string returned is something like that:
+        
+        in-hw-1-0-0-LPK25-MIDI-1
+        
+        or
+        
+        out-hw-2-0-0-MK-249C-USB-MIDI-keyboard-MIDI-
+
+        """
+        aliases=[_ffi.new("char[0xff]"), _ffi.new("char[0xff]")]
```
mgeier: You should probably use `port_name_size()` instead of `0xff`?

jofemodo: I think port_name_size() return the size of the "name", not the alias. I've not found any way of getting the alias size, so i decided to use 255, that should be enough ... i hope ;-)

jofemodo: Sorry, i just read the "set_alias" documentation and it says:

jofemodo:
I've implemented this feature because i need it. I've no need for the set_alias and unset_alias functions, although it's pretty simple to implement. I will do.

Also, i will return a list of aliases ;-)

jofemodo:
Done! ;-)

mgeier:
Thanks a lot for the changes!
Currently, I don't have time to review, I'll come back to you after the weekend.

jofemodo:
I just added a little test to examples for easier review ;-)

mgeier:
The API of `jack_port_get_aliases()` seems quite strange to me, because if it returns `1`, it doesn't tell us which of the two aliases is set ... or am I missing something?

Returning a list which may have empty strings in it (or even `None` values) seems a bit strange and un-Pythonic to me.
What about simply returning a "list of aliases" with 0, 1 or 2 elements, depending on the number of actual aliases?

mgeier:
Another thought: We could even return a `set` instead of a `list`?

jofemodo:
OK to the first part. Returning a set is not a good idea because the alias sorting is lost.

mgeier:
Thanks for the changes!

Does a "sorted list of aliases" even make sense?
Would you ever use the sorted-ness of the aliases in your code?
And if we return 1-element lists, we already give up on the sorted-ness.

How would you use the `aliases` property in your code?

Anyway, I'm not totally convinced about using a `set`. If you are convinced about using a `list`, I'm fine with it.

src/jack.py
```diff
@@ -1445,6 +1445,35 @@ def shortname(self, shortname):
...
 
+    @property
+    def aliases(self):
+        """A list of 2 strings containing the aliases for the JACK port."""
+        ctype="char[%d]" % _lib.jack_port_name_size()
+        aliases=[_ffi.new(ctype), _ffi.new(ctype)]
+        aliasesptr=_ffi.new("char *[]", aliases)
+        result=[]
+        if (_lib.jack_port_get_aliases(self._ptr, aliasesptr) > 0):
+          alias1=_ffi.string(aliases[0]).decode()
```
mgeier: You could reduce the duplication by using something like 

```python
for i in 0, 1:
    alias = ... aliases[i] ...
    if alias:
        result.append(alias)

**Comment by jofemodo** [comment]
OK! I've improved the code, reducing the redundance.
Regarding the list/set question, i prefer returning a list. In my code is not important the order, as only one alias is set, but i can imagine a case where you would like to get the "first" or "last" alias ;-)

File: src/jack.py [hunk 3]
```diff
@@ -1445,6 +1445,34 @@ def shortname(self, shortname):
         _check(_lib.jack_port_set_name(self._ptr, shortname.encode()),
                'Error setting port name')
 
+    @property
+    def aliases(self):
+        """A list of 2 strings containing the aliases for the JACK port."""
```
mgeier: Please updated the docstring to reflect the changed code.

src/jack.py
```diff
@@ -1445,6 +1445,34 @@ def shortname(self, shortname):
...
+        ctype="char[%d]" % _lib.jack_port_name_size()
+        aliases=[_ffi.new(ctype), _ffi.new(ctype)]
+        aliasesptr=_ffi.new("char *[]", aliases)
+        result=[]
+        if (_lib.jack_port_get_aliases(self._ptr, aliasesptr) > 0):
```
mgeier: The outer parentheses are not necessary.

The "> 0" isn't strictly necessary either, but you can keep it if you think it makes the situation clearer.

src/jack.py
```diff
@@ -1445,6 +1445,34 @@ def shortname(self, shortname):
         _check(_lib.jack_port_set_name(self._ptr, shortname.encode()),
                'Error setting port name')
 
+    @property
+    def aliases(self):
+        """A list of 2 strings containing the aliases for the JACK port."""
+        ctype="char[%d]" % _lib.jack_port_name_size()
```
mgeier: Please use `.format()`, because that's used in the rest of the code, too.

examples/test_alias.py
```diff
@@ -0,0 +1,52 @@
...
+    print('JACK server was started')
+else:
+    print('JACK server was already running')
+if client.status.name_not_unique:
+    print('unique client name generated:', client.name)
+
+ports = client.get_ports()
+
+print('Testing set_alias() ...')
+i=0
```
mgeier: Classic case for `enumerate()`: https://docs.python.org/3/library/functions.html#enumerate.

mgeier:
OK, thanks for the changes, let's keep the `list`.

Now there are just a few formal issues left, regarding [PEP-8](https://www.python.org/dev/peps/pep-0008/) and [PEP-257](https://www.python.org/dev/peps/pep-0257/).

```
$ python3 -m pycodestyle src/jack.py 
...
src/jack.py:1451:14: E225 missing whitespace around operator
src/jack.py:1452:16: E225 missing whitespace around operator
src/jack.py:1453:19: E225 missing whitespace around operator
src/jack.py:1454:15: E225 missing whitespace around operator
src/jack.py:1457:22: E225 missing whitespace around operator
src/jack.py:1465:67: W291 trailing whitespace
...
```

```
$ python3 -m pydocstyle src/jack.py
...
src/jack.py:1463 in public method `set_alias`:
        D205: 1 blank line required between summary line and description (found 0)
src/jack.py:1463 in public method `set_alias`:
        D209: Multi-line docstring closing quotes should be on a separate line
src/jack.py:1470 in public method `unset_alias`:
        D205: 1 blank line required between summary line and description (found 0)
src/jack.py:1470 in public method `unset_alias`:
        D209: Multi-line docstring closing quotes should be on a separate line
src/jack.py:1470 in public method `unset_alias`:
        D400: First line should end with a period (not ',')
...
```

Please rename the example file, because files starting and ending with "test" should be reserved for automated tests (e.g. py.test looks for those files).

jofemodo:
OK @mgeier! I hope everything is OK now ;-)

mgeier:
Thanks for the changes!

Did you consider these comments of mine?

* https://github.com/spatialaudio/jackclient-python/pull/45#discussion_r163813854
* https://github.com/spatialaudio/jackclient-python/pull/45#discussion_r163813654
* https://github.com/spatialaudio/jackclient-python/pull/45#discussion_r163812592

You are using tabs in the example file, which is a no-go. You should check for PEP-8 issues (e.g. with `python3 -m pycodestyle examples/aliases.py`). Ideally, you should use a plugin for your text editor to show those issues.

Finally, please also use `.format()` in the example.

jofemodo:
OK! Let's see now ... cross my fingers ;-D

mgeier:
Thanks a lot @jofemodo!

jofemodo:
It's been a pleasure, @mgeier  ;-)

Issue #40, opened 2017-05-28
Support for the metadata API?


simonvanderveldt:
First of all, thanks for this package! Helps a lot with neatly automating my setup :)

I was wondering if [Jack's metadata API](http://jackaudio.org/api/group__Metadata.html) can be used through jackclient-python?
I couldn't find any references to it in the the current code, so maybe some work needs to be done to enable it? If so, what would that be? Maybe I can be of help.

mgeier:
TBH, I wasn't aware of this API and therefore it's currently not supported.
But I don't see a reason why not to implement this.

Before thinking about the implementation, we should think about the potential API on the Python side.
Could you please make some suggestions how this could look like?
It would be nice to see some hypothetical code examples for this.

simonvanderveldt:
@mgeier I'm prototyping a way to treat separate jack clients as a group, i.e. create them, create the correct connections and present them to the user as a "stack", similar to how for example [Ardour](https://ardour.org/images/lsd-mixer.png) or [non-mixer](http://non.tuxfamily.org/mixer/doc/non-mixer-complex.png) s
how plugins.

To be able to do this I was looking at a way to add some metadata/tags to the clients I spawn and low and behold there's metadata API in jack :)

The jack client based method's as they are available in the Jack API look good enough to me, though it would obviously be nice to have some higher lever/object oriented method's as well, for example to allow adding the metadata when creating a client.

mgeier:
I don't really understand why some functions take a `jack_client_t*` and others don't.
I was thinking about creating new methods on `jack.Client` objects, but if not all functions take a `jack_client_t*`, that doesn't really make sense, does it?

I also don't really understand how different data types are supposed to be handled, nor do I understand what a "subject" is supposed to be.

Can you please provide some concrete (but still hypothetical) examples that could help clarify this?

simonvanderveldt:
> I don't really understand why some functions take a jack_client_t* and others don't.
> I was thinking about creating new methods on jack.Client objects, but if not all functions take a jack_client_t*, that doesn't really make sense, does it?

Looking at the API docs, apparently getters don't need a `jack_client_t`, whereas all setters do.
As to why that is I don't know. Probably easiest to ask in #lad, #jack or on the lad mailing list?

> I also don't really understand how different data types are supposed to be handled, nor do I understand what a "subject" is supposed to be.

If I understand it correctly the subject is the client you want to set/update the metadata for, identified by its UUID.

[...]

PR #64, opened 2019-04-27
Add support for metadata API

mgeier:
See #40.

### DIFF
--- a/src/jack.py
+++ b/src/jack.py
@@ -2550,6 +2550,9 @@ class CallbackExit(Exception):
 
 
 def _uuid_parse(uuid):
+    if isinstance(uuid, (Client, Port)):
+        uuid = uuid.uuid
+
     if isinstance(uuid, int):
         return uuid
     elif isinstance(uuid, str):
--- a/src/jack.py
+++ b/src/jack.py
@@ -146,7 +146,7 @@ class Client(object):
     @property
     def name(self):
         """The name of the JACK client (read-only)."""
-        return _ffi.string(_lib.jack_get_client_name(self._ptr)).decode()
+        return _decode(_lib.jack_get_client_name(self._ptr))
 
     @property
     def samplerate(self):
@@ -605,7 +605,7 @@ class Client(object):
         """
         @self._callback('JackInfoShutdownCallback')
         def callback_wrapper(code, reason, _):
-            callback(Status(code), _ffi.string(reason).decode())
+            callback(Status(code), _decode(reason))
 
         _lib.jack_on_info_shutdown(self._ptr, callback_wrapper, _ffi.NULL)
 
@@ -844,7 +844,7 @@ class Client(object):
         """
         @self._callback('JackClientRegistrationCallback')
         def callback_wrapper(name, register, _):
-            callback(_ffi.string(name).decode(), bool(register))
+            callback(_decode(name), bool(register))
 
         _check(_lib.jack_set_client_registration_callback(
             self._ptr, callback_wrapper, _ffi.NULL),
@@ -1055,8 +1055,7 @@ class Client(object):
             else:
                 port = None
             try:
-                callback(port, _ffi.string(old_name).decode(),
-                         _ffi.string(new_name).decode())
+                callback(port, _decode(old_name), _decode(new_name))
             except CallbackExit:
                 return _FAILURE
             return _SUCCESS
@@ -1237,7 +1236,7 @@ class Client(object):
             self._ptr, name.encode()), _lib.jack_free)
         if not uuid:
             raise JackError('Unable to get session ID for {0!r}'.format(name))
-        return _ffi.string(uuid).decode()
+        return _decode(uuid)
 
     def get_client_name_by_uuid(self, uuid):
         """Get the client name for a session ID.
@@ -1250,7 +1249,7 @@ class Client(object):
             self._ptr, uuid.encode()), _lib.jack_free)
         if not name:
             raise JackError('Unable to get client name for {0!r}'.format(uuid))
-        return _ffi.string(name).decode()
+        return _decode(name)
 
     def get_port_by_name(self, name):
         """Get port by name.
@@ -1358,7 +1357,7 @@ class Client(object):
                 name = names[idx]
                 if not name:
                     break
-                ports.append(self.get_port_by_name(_ffi.string(name).decode()))
+                ports.append(self.get_port_by_name(_decode(name)))
                 idx += 1
         return ports
 
@@ -1425,7 +1424,7 @@ class Port(object):
     @property
     def name(self):
         """Full name of the JACK port (read-only)."""
-        return _ffi.string(_lib.jack_port_name(self._ptr)).decode()
+        return _decode(_lib.jack_port_name(self._ptr))
 
     @property
     def shortname(self):
@@ -1438,7 +1437,7 @@ class Port(object):
         `port_name_size()`, it will be truncated.
 
         """
-        return _ffi.string(_lib.jack_port_short_name(self._ptr)).decode()
+        return _decode(_lib.jack_port_short_name(self._ptr))
 
     @shortname.setter
     def shortname(self, shortname):
@@ -1454,7 +1453,7 @@ class Port(object):
         result = []
         if _lib.jack_port_get_aliases(self._ptr, aliasesptr) > 0:
             for i in 0, 1:
-                alias = _ffi.string(aliases[i]).decode()
+                alias = _decode(aliases[i])
                 if alias:
                     result.append(alias)
 
@@ -2385,7 +2384,7 @@ def version():
 
 def version_string():
     """Get human-readable JACK version."""
-    return _ffi.string(_lib.jack_get_version_string()).decode()
+    return _decode(_lib.jack_get_version_string())
 
 
 def client_name_size():
@@ -2460,7 +2459,7 @@ def _set_error_or_info_function(callback, setter):
     else:
         @_ffi.callback('void (*)(const char*)')
         def callback_wrapper(msg):
-            callback(_ffi.string(msg).decode())
+            callback(_decode(msg))
 
         _keepalive[setter] = callback_wrapper
     setter(callback_wrapper)
@@ -2473,3 +2472,7 @@ def _check(error_code, msg):
     """Check error code and raise JackError if non-zero."""
     if error_code:
         raise JackError('{0} ({1})'.format(msg, error_code))
+
+
+def _decode(cdata):
+    return _ffi.string(cdata).decode()
--- a/src/jack.py
+++ b/src/jack.py
@@ -149,6 +149,14 @@ class Client(object):
         return _decode(_lib.jack_get_client_name(self._ptr))
 
     @property
+    def uuid(self):
+        """The UUID of the JACK client (read-only)."""
+        uuid = _ffi.gc(_lib.jack_client_get_uuid(self._ptr), _lib.jack_free)
+        if not uuid:
+            raise JackError('Unable to get UUID')
+        return _ffi.string(uuid).decode()
+
+    @property
     def samplerate(self):
         """The sample rate of the JACK system (read-only)."""
         return _lib.jack_get_sample_rate(self._ptr)
--- a/src/jack.py
+++ b/src/jack.py
@@ -74,7 +74,10 @@ def _decode(cdata):
 # Get metadata constants from library
 for name in dir(_lib):
     if name.startswith('JACK_METADATA_'):
-        globals()[name[5:]] = _decode(getattr(_lib, name))
+        try:
+            globals()[name[5:]] = _decode(getattr(_lib, name))
+        except _ffi.error:
+            pass
 else:
     del name
 
--- a/src/jack.py
+++ b/src/jack.py
@@ -56,8 +56,27 @@ STARTING = _lib.JackTransportStarting
 NETSTARTING = _lib.JackTransportNetStarting
 """Waiting for sync ready on the network."""
 
+PROPERTY_CREATED = _lib.PropertyCreated
+"""A property was created.  See `Client.set_property_change_callback()`."""
+PROPERTY_CHANGED = _lib.PropertyChanged
+"""A property was changed.  See `Client.set_property_change_callback()`."""
+PROPERTY_DELETED = _lib.PropertyDeleted
+"""A property was deleted.  See `Client.set_property_change_callback()`."""
+
 _SUCCESS = 0
 _FAILURE = 1
+
+
+def _decode(cdata):
+    return _ffi.string(cdata).decode()
+
+
+# Get metadata constants from library
+for name in dir(_lib):
+    if name.startswith('JACK_METADATA_'):
+        globals()[name[5:]] = _decode(getattr(_lib, name))
+else:
+    del name
 
 
 class Client(object):
@@ -154,7 +173,7 @@ class Client(object):
         uuid = _ffi.gc(_lib.jack_client_get_uuid(self._ptr), _lib.jack_free)
         if not uuid:
             raise JackError('Unable to get UUID')
-        return _ffi.string(uuid).decode()
+        return _decode(uuid)
 
     @property
     def samplerate(self):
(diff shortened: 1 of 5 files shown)


src/jack.py
```diff
@@ -1328,6 +1383,141 @@ def get_ports(self, name_pattern='', is_audio=False, is_midi=False,
...
+
+        See Also
+        --------
+        set_property
+        get_property
+        get_properties
+        get_all_properties
+        remove_properties
+        remove_all_properties
+        set_property_change_callback
```
SpotlightKid: Should these be marked up with backticks?

src/jack.py
```diff
@@ -1328,6 +1383,141 @@ def get_ports(self, name_pattern='', is_audio=False, is_midi=False,
...
+        get_properties
+        get_all_properties
+        remove_property
+        remove_properties
+        remove_all_properties
+        set_property_change_callback
+
+        """
+        subject = _uuid_parse(subject)
+        if isinstance(value, str):
```
SpotlightKid: This is problematic on Python 2. The `str` type is essentially the same as `bytes` on Python 3 there. It's value should not be encoded, it already is. `unicode` instances, on the other hand, should be encoded. One solution is to use `isinstance(s, six.text_type)` or have something like this at the top of your module:

    if bytes is str:
        text_type = unicode
    else:
        text_type = str

(But I'm not sure whether the latter works on Python 2.6.)

src/jack.py
```diff
@@ -1328,6 +1383,141 @@ def get_ports(self, name_pattern='', is_audio=False, is_midi=False,
...
+            value = value.encode()
+        if isinstance(type, str):
```
SpotlightKid: Same problem with Python 2 here, but in this instance it shouldn't matter, because type should be ASCII only.

src/jack.py
```diff
@@ -2356,6 +2546,140 @@ class CallbackExit(Exception):
     pass
 
 
+def _uuid_parse(uuid):
+    if isinstance(uuid, int):
+        return uuid
+    elif isinstance(uuid, str):
```
SpotlightKid: This would not accept `unicode` strings on Python 2.

Maybe check `isinstance(uuid, (bytes, text_type))` here and only encode uuid, if `isinstance(uuid, text_type)`.

src/jack.py
```diff
@@ -1328,6 +1383,141 @@ def get_ports(self, name_pattern='', is_audio=False, is_midi=False,
...
+
+        See Also
+        --------
+        set_property
+        get_property
+        get_properties
+        get_all_properties
+        remove_property
+        remove_all_properties
+        set_property_change_callback
```
SpotlightKid: Should these be marked up with backticks?

src/jack.py
```diff
@@ -1328,6 +1383,141 @@ def get_ports(self, name_pattern='', is_audio=False, is_midi=False,
...
+
+        See Also
+        --------
+        set_property
+        get_property
+        get_properties
+        get_all_properties
+        remove_property
+        remove_properties
+        set_property_change_callback
```
SpotlightKid: Should these be marked up with backticks?

src/jack.py
```diff
@@ -2356,6 +2546,140 @@ class CallbackExit(Exception):
...
+
+    See Also
+    --------
+    Client.set_property
+    get_properties
+    get_all_properties
+    Client.remove_property
+    Client.remove_properties
+    Client.remove_all_properties
+    set_property_change_callback
```
SpotlightKid: Should these be marked up with backticks?

src/jack.py
```diff
@@ -2356,6 +2546,140 @@ class CallbackExit(Exception):
...
+
+    See Also
+    --------
+    Client.set_property
+    get_property
+    get_all_properties
+    Client.remove_property
+    Client.remove_properties
+    Client.remove_all_properties
+    set_property_change_callback
```
SpotlightKid: Should these be marked up with backticks?

src/jack.py
```diff
@@ -56,10 +56,29 @@
...
+"""A property was changed.  See `Client.set_property_change_callback()`."""
+PROPERTY_DELETED = _lib.PropertyDeleted
+"""A property was deleted.  See `Client.set_property_change_callback()`."""
+
 _SUCCESS = 0
 _FAILURE = 1
 
 
+def _decode(cdata):
+    return _ffi.string(cdata).decode()
```
SpotlightKid: This handles only ASCII strings, doesn't it?

Maybe allow to pass an encoding?

```
def _decode(cdata, encoding='utf-8'):
    return _ffi.string(cdata).decode(encoding)
```

src/jack.py
```diff
@@ -56,10 +56,29 @@
...
+
+
+# Get metadata constants from library
+for name in dir(_lib):
+    if name.startswith('JACK_METADATA_'):
+        globals()[name[5:]] = _decode(getattr(_lib, name))
```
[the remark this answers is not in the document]
mgeier: Thanks, that's a great trick!

However, I think i would like to be a bit more defensive in case those constants are not defined in the library. I wanted to do that at a later point, but I guess I can already do it now: 5d9b56541aa9659ecb87ea14b108574b0beb55bb

src/jack.py
```diff
@@ -1328,6 +1383,141 @@ def get_ports(self, name_pattern='', is_audio=False, is_midi=False,
...
+
+        See Also
+        --------
+        get_property
+        get_properties
+        get_all_properties
+        remove_property
+        remove_properties
+        remove_all_properties
+        set_property_change_callback
```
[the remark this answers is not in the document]
mgeier: Those are converted to links correctly, so I don't think that's necessary. I guess the `sphinx.ext.napoleon` extension takes care of that.

mgeier:
@SpotlightKid Thanks for the review!

I don't have any motivation to actively write code for Python 2 support anymore. This PR assumes Python 3.

Do you need Python 2 support?

If yes, could you please make a new PR based on my PR that takes care of Python 2?

src/jack.py
mgeier: I've used this in the past:

```python
try:
    _unicode = unicode  # doesn't exist in Python 3.x
except NameError:
    _unicode = str
```

I think adding a dependency on `six` is overkill here.

src/jack.py
mgeier: I was assuming Python 3, where `encoding="utf-8", errors="strict"` is the default.

I guess for Python 2 support we would have to make this explicit. But I don't think we have to make it configurable.

SpotlightKid:
No, I myself decided to drop Python 2 support in my own projects successivley as well. It's just that the readme still mentions even Python 2.6 support :-).

src/jack.py
SpotlightKid: The default can still be changed, even in Python 3, though.

mgeier:
@SpotlightKid That's a good point! To fix this, I've just created #65, which I'm planning to merge before this PR.

Assuming only Python 3 compatibility, are there still open issues?

SpotlightKid:
> Any suggestions for API improvements?

1. Should the default value for `type` really be `''` not `None`?
2. You could add convenience methods to the `Client` and `Port` classes:

    * `Client.get_property(key)`
    * `Client.get_properties()`
    * `Port.get_property(key)`
    * `Port.get_properties()`
    * `Port.set_property(key, type=None)`
    * `Port.remove_property(key)`

    The problem is, that the following would conflict with the methods to set/remove properties on other subjects:

    * `Client.set_property(key, value, type=None)`
    * `Client.remove_property(key)`

   You could make all parameters keyword parameters and make `subject` optional, with a default of `None`, which means, replace it with `self.uuid`. But it could be confusing and it would mean all other parameters would have to be passed as keyword args.

3. You could allow the `subject` argument value for all metadata functions and methods to be also a `Client` or `Port` instance. Just put the following at the start of the `uuid_parse` function with:

        if isinstance(uuid, (Client, Port):
            uuid = uuid.uuid

4. I think returning the property values as byte strings is a bit inconvenient. Here's how I would handle it:

```diff
diff --git a/src/jack.py b/src/jack.py
index 6313e05..b41db6f 100644
--- a/src/jack.py
+++ b/src/jack.py
@@ -2560,21 +2560,29 @@ def _uuid_parse(uuid):
     raise TypeError('Invalid UUID: {!r}'.format(uuid))
 
 
-def _description_to_dict(desc):
+def _decode_property(prop, encoding='utf-8'):
+    type_ = _decode(prop.type) if prop.type else None
+    value = _ffi.string(prop.data)
+
+    if not type_ or type_.startswith('text/') and encoding:
+        value = value.decode(encoding, errors='ignore')
+
+    return value, type_
+
+
+def _description_to_dict(desc, encoding):
     assert desc != _ffi.NULL
     prop_dict = {}
     for i in range(desc.property_cnt):
         prop = desc.properties[i]
         key = _decode(prop.key)
-        prop_dict[key] = (
-            _ffi.string(prop.data),
-            _decode(prop.type) if prop.type else '')
+        prop_dict[key] = _decode_property(prop, encoding)
     free_description_itself = 0
     _lib.jack_free_description(desc, free_description_itself)
     return prop_dict
 
 
-def get_property(subject, key):
+def get_property(subject, key, encoding='utf-8'):
     """Get a metadata property on *subject*.
 
     Parameters
@@ -2610,11 +2618,17 @@ def get_property(subject, key):
     if _lib.jack_get_property(subject, key.encode(), data, type) != 0:
         return None
     data = _ffi.gc(data[0], _lib.jack_free)
+    data = _ffi.string(data)
     type = _ffi.gc(type[0], _lib.jack_free)
-    return _ffi.string(data), _decode(type) if type else ''
+    type = _decode(type) if type else None
+
+    if not type or type.startswith('text/') and encoding:
+        data = data.decode(encoding, errors='ignore')
+
+    return data, type
 
 
-def get_properties(subject):
+def get_properties(subject, encoding='utf8'):
     """Get all metadata properties of *subject*.
 
     Parameters
@@ -2647,10 +2661,10 @@ def get_properties(subject):
         raise RuntimeError(
             'Unable to get properties for subject {!r}'.format(subject))
     assert number == desc.property_cnt
-    return _description_to_dict(desc)
+    return _description_to_dict(desc, encoding)
 
 
-def get_all_properties():
+def get_all_properties(encoding='utf8'):
     """Get all properties for all subjects with metadata.
 
     Returns
@@ -2679,7 +2693,7 @@ def get_all_properties():
     for idx in range(number):
         desc = descs + idx
         assert desc.subject not in prop_dict
-        prop_dict[desc.subject] = _description_to_dict(desc)
+        prop_dict[desc.subject] = _description_to_dict(desc, encoding)
     return prop_dict
 ```
    You could still pass`encoding=None` to get a `bytes` value.

mgeier:
Thanks a lot for those suggestions!

> Should the default value for `type` really be `''` not `None`?

I was considering this, but even when I was passing `NULL` as "type" argument to `jack_set_property()`, I didn't get `NULL` out from `jack_get_property()`.

Probably I did something wrong. Did you try this?

> You could add convenience methods to the `Client` and `Port` classes

You already mentioned some problems with this. Since I need a `Client` object for creating/deleting *any* property, it would be confusing to have one method that acts on `self` and a similarly named method that acts on a given "subject".

I think it is more consistent if the "subject" always has to be explicit.

Yet another problem is that changing properties on `Port` objects would not work, because they don't have an associated client. This would only work with `OwnPort` objects.

> You could allow the `subject` argument value for all metadata functions and methods to be also a `Client` or `Port` instance.

Good idea, I've added your suggestion in d79be4a118acad1abcd02dc0e2d6c3afc0838aba.

> I think returning the property values as byte strings is a bit inconvenient.

That's true, but I think `encoding=None` would not be understood as disabling the `decode()` step, but rather as using the default encoding, which is UTF-8.

Also, we shouldn't expose an `encoding` parameter without also having an `errors` parameter.

I don't really like the `type.startswith('text/')` heuristic in such a low-level module. I think such a thing should only be used in user code or probably in a wrapper module with more specific use cases.

Finally, I'm skeptical whether it is a good idea to return different types from a function depending on external factors.

The JACK docs state that when `type` is empty, the data is expected to be UTF-8 encoded. I was thinking about returning one of two types:

* `(b'some data', 'some type')`
* `'a Unicode string'`

But then users wouldn't be able to use tuple unpacking and I think the handling would be more complicated. I think this is easier to handle:

* `(b'some data', 'some type')`
* `(b'some UTF-8 encoded data', '')`

If I understood correctly, your suggestion would return one of those:

* `(b'some data', 'some type')`
* `('a Unicode string', 'text/something')`
* `('a Unicode string', '')`

Wouldn't this make user code more complicated because they always have to check if they got a `str` or `bytes` back from the function call?

SpotlightKid:
> but even when I was passing NULL as "type" argument to jack_set_property(), I didn't get NULL out from jack_get_property(). [...] Did you try this?

Yes, when using ctypes, I get a null pointer back, when the type is not set. I think that's just an implementation detail of how cffi works?

SpotlightKid:
I concur with the rest of your points, though having some more user-friendly wrapper code would probably be nice.

mgeier:
> Yes, when using ctypes, I get a null pointer back, when the type is not set.

OK, I tried it again, and it turns out I was mistaken about getting `NULL` back from `jack_get_property()`.

If I set `NULL`, I get `NULL`, that works fine.
But if I set an empty string, I don't get an empty string back (instead I get `NULL`).

I think that was the reason why I chose to not support `type=None`.

Are you getting an empty string back?

> having some more user-friendly wrapper code would probably be nice.

I agree, but I think adding this to the `Client` class would be more confusing than helpful.

I think I would prefer providing a very basic low-level API, and users can create their own wrappers based on their specific use cases.

SpotlightKid:
> I think that was the reason why I chose to not support type=None.

Huh? I don't understand the logic behind that. Anyway, setting `type` to an empty string would be against the spec, since it should be either a mime-type or a URI.

mgeier:
> > I think that was the reason why I chose to not support `type=None`.
>
> Huh? I don't understand the logic behind that.

If I *set* the `type` to `''`, I would expect to *get* a `type` of `''` back, but instead I get `NULL` (or `None`, when translated to Python).

If I allow both `type=None` and `type=''`, one of them will behave inconsistently.

Therefore, I think it's better to allow only one of them.

Since disallowing `type=''` would be really strange (wouldn't it?), I think it makes much more sense to disallow `type=None`.

Does that argument make sense?

> Anyway, setting `type` to an empty string would be against the spec, since it should be either a mime-type or a URI.

According to http://jackaudio.org/api/structjack__property__t.html#a298b7c97464c5ac3a6070c467732bd9c:

"If type is NULL or **empty**, the data is assumed to be a UTF-8 encoded string (text/plain)." (emphasis mine)

Do you have a different spec?

SpotlightKid:
> Do you have a different spec?

No, I just didn't interpret "empty" as an empty string ;)

> If I allow both type=None and type='', one of them will behave inconsistently.

OTOH, I would find it surprising that if no `type` argument is given to `set_property` it will result in `type` being set to an empty string.

> If I set the type to '', I would expect to get a type of '' back, but instead I get NULL

Which strengthens my belief that the intention of the JACK developers was, that if type is set, it should be a string containing something (a mime-type or a URI).

Maybe this should be raised as an issue with JACK? Have you tested the behavior of Jack2 (current git)?

mgeier:
> OTOH, I would find it surprising that if no `type` argument is given to `set_property` it will result in `type` being set to an empty string.

The function signature of `set_property()` clearly shows `type=''`.

There is no reason for a user to believe that `type` should ever be anything other than a `str`.

Would you expect `None`?
What would make you expect that?

>> If I set the type to '', I would expect to get a type of '' back, but instead I get NULL
>
>Which strengthens my belief that the intention of the JACK developers was, that if type is set, it should be a string containing something (a mime-type or a URI).

I don't think so.

I think that's just them dealing with an idiosyncrasy of C, where a "string" can be either a `NULL` pointer or a pointer to some zero-terminated characters.

There is no reason to provide two different representations for "empty" (`NULL` pointer and pointer to `'\0'`), but C forces them to handle both cases.

No other language has to worry about those two cases and can simply use the empty string to signify "empty".

> Maybe this should be raised as an issue with JACK?

I don't think that's an issue. But feel free to open an issue there if you think there is one!

> Have you tested the behavior of Jack2 (current git)?

No, I didn't. But whatever its behavior is, I think my current solution would work (and return `''` for empty types).

SpotlightKid:
Sorry for not having answered yet.

I think we we both made our views clear now and it's up to you to decide what the API should look like in the end. I'm looking forward to using it.

SpotlightKid:
Just a little ping... anything still preventing this from being merged?

mgeier:
@SpotlightKid Sorry for the late response and thanks for the reminder!

I've just merged the current state, but if you have further suggestions for changes feel free to comment here or make a new PR!

PR #65, opened 2019-04-30
Officially drop Python 2 compatibility

mgeier:


setup.py
```diff
@@ -30,7 +30,6 @@
         'License :: OSI Approved :: MIT License',
         'Operating System :: OS Independent',
         'Programming Language :: Python',
-        'Programming Language :: Python :: 2',
         'Programming Language :: Python :: 3',
```
SpotlightKid: You could add `'Programming Language :: Python :: 3 :: Only'` here.

mgeier: Done.

PR #71, opened 2019-09-03
Make TransportState class hashable

SpotlightKid:
This allows instances to be used as dict keys, without breaking encapsulation by accessing the internal attribute `_code` directly.


mgeier:
Thanks!

PR #73, opened 2019-09-03
Add 'sphinx_rtd_theme' to doc requirements

SpotlightKid:
With newer Sphinx versions you get the following error on `python setup.py build_sphinx`, if `sphinx_rtd_theme` is not installed explicitly:

```
Theme error:
sphinx_rtd_theme is no longer a hard dependency since version 1.4.0. Please install it manually.(pip install sphinx_rtd_theme)
```

### DIFF
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -1,3 +1,4 @@
 Jinja2
 Pygments
 Sphinx>=1.3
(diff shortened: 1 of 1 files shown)


mgeier:
Thanks!

PR #72, opened 2019-09-03
Add 'release_timebase' method to Client class

SpotlightKid:
Signed-off-by: Christopher Arndt <chris@chrisarndt.de>

PR #74, opened 2019-09-11
Add examples for for JACK transport and timebase master

SpotlightKid:
Follow up to #72. Obsoletes #19.

### DIFF
--- a/examples/timebase_master.py
+++ b/examples/timebase_master.py
@@ -105,7 +105,7 @@ def main(args=None):
         default=120.0,
         help="Tempo in beats per minute (0.1-300.0, default: %(default)s)")
 
-    args = ap.parse_args(args if args is not None else sys.argv[1:])
+    args = ap.parse_args(args)
 
     try:
         beats_per_bar, beat_type = (int(x) for x in args.meter.split('/', 1))
--- a/examples/timebase_master.py
+++ b/examples/timebase_master.py
@@ -36,7 +36,7 @@ class TimebaseMasterClient(jack.Client):
             pos.beats_per_minute = self.bpm
             pos.beat_type = float(self.beat_type)
             pos.ticks_per_beat = float(self.ticks_per_beat)
-            pos.valid |= jack._lib.JackPositionBBT
+            pos.valid = jack._lib.JackPositionBBT
 
             minutes = pos.frame / (pos.frame_rate * 60.0)
             abs_tick = minutes * self.bpm * self.ticks_per_beat
--- a/examples/timebase_master.py
+++ b/examples/timebase_master.py
@@ -140,4 +140,4 @@ def main(args=None):
 
 
 if __name__ == '__main__':
-    sys.exit(main(sys.argv[1:]) or 0)
+    sys.exit(main() or 0)
--- a/examples/timebase_master.py
+++ b/examples/timebase_master.py
@@ -1,28 +1,32 @@
-#!/usr/bin/env python
-# -*- coding: utf-8 -*-
+#!/usr/bin/env python3
 #
-#  timebase_masster.py
+#  timebase_master.py
 #
 """A simple JACK timebase master."""
 
 import argparse
 import sys
-import time
+from threading import Event
 
 import jack
 
 
 class TimebaseMasterClient(jack.Client):
-    def __init__(self, name, bpm=120.0, meter='4/4', beats_per_bar=4,
-                 beat_type=4, ticks_per_beat=1920, conditional=False,
-                 debug=False, **kw):
-        self.beats_per_bar = beats_per_bar
-        self.beat_type = beat_type
+    def __init__(self, name, *, bpm=120.0, beats_per_bar=4, beat_type=4,
+                 ticks_per_beat=1920, conditional=False, debug=False, **kw):
+        super().__init__(name, **kw)
+        self.beats_per_bar = int(beats_per_bar)
+        self.beat_type = int(beat_type)
         self.bpm = bpm
         self.conditional = conditional
         self.debug = debug
         self.ticks_per_beat = int(ticks_per_beat)
-        super().__init__(name, **kw)
+        self.stop_event = Event()
+        self.set_shutdown_callback(self.shutdown)
+
+    def shutdown(self, status, reason):
+        print('JACK shutdown:', reason, status)
+        self.stop_event.set()
 
     def _tb_callback(self, state, nframes, pos, new_pos):
         if self.debug and new_pos:
@@ -31,20 +35,20 @@ class TimebaseMasterClient(jack.Client):
         # Adapted from:
         # https://github.com/jackaudio/jack2/blob/develop/example-clients/transport.c#L66
         if new_pos:
-            pos.beats_per_bar = self.beats_per_bar
+            pos.beats_per_bar = float(self.beats_per_bar)
             pos.beats_per_minute = self.bpm
-            pos.beat_type = self.beat_type
-            pos.ticks_per_beat = self.ticks_per_beat
+            pos.beat_type = float(self.beat_type)
+            pos.ticks_per_beat = float(self.ticks_per_beat)
             pos.valid |= jack._lib.JackPositionBBT
 
             minutes = pos.frame / (pos.frame_rate * 60.0)
-            abs_tick = minutes * pos.beats_per_minute * pos.ticks_per_beat
+            abs_tick = minutes * self.bpm * self.ticks_per_beat
             abs_beat = abs_tick / self.ticks_per_beat
 
-            pos.bar = int(abs_beat / pos.beats_per_bar)
-            pos.beat = int(abs_beat - (pos.bar * pos.beats_per_bar) + 1)
-            pos.tick = int(abs_tick - (abs_beat * pos.ticks_per_beat))
-            pos.bar_start_tick = pos.bar * pos.beats_per_bar * pos.ticks_per_beat
+            pos.bar = int(abs_beat / self.beats_per_bar)
+            pos.beat = int(abs_beat - (pos.bar * self.beats_per_bar) + 1)
+            pos.tick = int(abs_tick - (abs_beat * self.ticks_per_beat))
+            pos.bar_start_tick = pos.bar * self.beats_per_bar * self.ticks_per_beat
             pos.bar += 1  # adjust start to bar 1
         else:
             # Compute BBT info based on previous period.
@@ -73,16 +77,16 @@ def main(args=None):
     ap = argparse.ArgumentParser(description=__doc__)
     ap.add_argument(
         '-d', '--debug',
-        action="store_true",
+        action='store_true',
         help="Enable debug messages")
     ap.add_argument(
         '-c', '--conditional',
-        action="store_true",
+        action='store_true',
         help="Exit if another timebase master is already active")
     ap.add_argument(
         '-n', '--client-name',
         metavar='NAME',
-        default="timebase",
+        default='timebase',
         help="JACK client name (default: %(default)s)")
     ap.add_argument(
         '-m', '--meter',
@@ -99,9 +103,9 @@ def main(args=None):
         nargs='?',
         type=float,
         default=120.0,
-        help="Tempo in beats per minute (default: %(default)s)")
+        help="Tempo in beats per minute (0.1-300.0, default: %(default)s)")
 
-    args = ap.parse_args(args if args is not None else sys.argsv[1:])
+    args = ap.parse_args(args if args is not None else sys.argv[1:])
 
     try:
         beats_per_bar, beat_type = (int(x) for x in args.meter.split('/', 1))
@@ -113,32 +117,29 @@ def main(args=None):
     try:
         tbmaster = TimebaseMasterClient(
             args.client_name,
-            bpm=max(0.0, min(300.0, args.tempo)),
+            bpm=max(0.1, min(300.0, args.tempo)),
             beats_per_bar=beats_per_bar,
             beat_type=beat_type,
             ticks_per_beat=args.ticks_per_beat,
             debug=args.debug)
-        tbmaster.activate()
     except jack.JackError as exc:
         return "Could not create timebase master JACK client: %s" % exc
 
-    if tbmaster.become_timebase_master(args.conditional):
-        try:
-            while True:
-                time.sleep(0.1)
-        except KeyboardInterrupt:
-            print("\nBye!")
-        finally:
+    with tbmaster:
+        if tbmaster.become_timebase_master(args.conditional):
             try:
-                tbmaster.release_timebase()
-            except jack.JackError:
-                # another JACK client might have grabbed timebase master
-                pass
-    else:
-        print("Timebase master already present. Exiting...")
-
-    tbmaster.deactivate()
-    tbmaster.close()
+                print("Press Ctrl-C to quit...")
+                tbmaster.stop_event.wait()
+            except KeyboardInterrupt:
+                print('')
+            finally:
+                try:
+                    tbmaster.release_timebase()
+                except jack.JackError:
+                    # another JACK client might have grabbed timebase master
+                    pass
+        else:
+            return "Timebase master already present. Exiting..."
 
 
 if __name__ == '__main__':
--- a/examples/timebase_master.py
+++ b/examples/timebase_master.py
@@ -1,0 +1,145 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#
+#  timebase_masster.py
+#
+"""A simple JACK timebase master."""
+
+import argparse
+import sys
+import time
+
+import jack
+
+
+class TimebaseMasterClient(jack.Client):
+    def __init__(self, name, bpm=120.0, meter='4/4', beats_per_bar=4,
+                 beat_type=4, ticks_per_beat=1920, conditional=False,
+                 debug=False, **kw):
+        self.beats_per_bar = beats_per_bar
+        self.beat_type = beat_type
+        self.bpm = bpm
+        self.conditional = conditional
+        self.debug = debug
+        self.ticks_per_beat = int(ticks_per_beat)
+        super().__init__(name, **kw)
+
+    def _tb_callback(self, state, nframes, pos, new_pos):
+        if self.debug and new_pos:
+            print("New pos:", jack.position2dict(pos))
+
+        # Adapted from:
+        # https://github.com/jackaudio/jack2/blob/develop/example-clients/transport.c#L66
+        if new_pos:
+            pos.beats_per_bar = self.beats_per_bar
+            pos.beats_per_minute = self.bpm
+            pos.beat_type = self.beat_type
+            pos.ticks_per_beat = self.ticks_per_beat
+            pos.valid |= jack._lib.JackPositionBBT
+
+            minutes = pos.frame / (pos.frame_rate * 60.0)
+            abs_tick = minutes * pos.beats_per_minute * pos.ticks_per_beat
+            abs_beat = abs_tick / self.ticks_per_beat
+
+            pos.bar = int(abs_beat / pos.beats_per_bar)
+            pos.beat = int(abs_beat - (pos.bar * pos.beats_per_bar) + 1)
+            pos.tick = int(abs_tick - (abs_beat * pos.ticks_per_beat))
+            pos.bar_start_tick = pos.bar * pos.beats_per_bar * pos.ticks_per_beat
+            pos.bar += 1  # adjust start to bar 1
+        else:
+            # Compute BBT info based on previous period.
+            pos.tick += int(nframes * pos.ticks_per_beat *
+                            pos.beats_per_minute / (pos.frame_rate * 60))
+
+            while pos.tick >= pos.ticks_per_beat:
+                pos.tick -= int(pos.ticks_per_beat)
+                pos.beat += 1
+
+                if pos.beat > pos.beats_per_bar:
+                    pos.beat = 1
+                    pos.bar += 1
+                    pos.bar_start_tick += pos.beats_per_bar * pos.ticks_per_beat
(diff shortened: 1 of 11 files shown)


examples/timebase_master.py
```diff
@@ -0,0 +1,148 @@
...
+        default=1920,
+        help="Ticks per beat (default: %(default)s)")
+    ap.add_argument(
+        'tempo',
+        nargs='?',
+        type=float,
+        default=120.0,
+        help="Tempo in beats per minute (default: %(default)s)")
+
+    args = ap.parse_args(args if args is not None else sys.argv[1:])
```
mgeier: I think you are doing what Python would be doing anyway.

https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.parse_args

I think you can just pass `args` and be done with it.

examples/timebase_master.py
```diff
@@ -0,0 +1,148 @@
...
+
+    try:
+        beats_per_bar, beat_type = (int(x) for x in args.meter.split('/', 1))
```
mgeier: The second argument to `split()` is not necessary, but I guess it doesn't hurt?

examples/timebase_master.py
```diff
@@ -0,0 +1,148 @@
...
+    except (TypeError, ValueError):
+        print("Error: invalid meter: %s\n" % args.meter)
```
mgeier: I think I normally use `.format()`, is there a reason to use the old-school string interpolation?

examples/timebase_master.py
```diff
@@ -0,0 +1,148 @@
...
+        ap.print_help()
+        return 2
```
mgeier: I would use `ap.error(...)` for this.

https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.error

examples/timebase_master.py
```diff
@@ -0,0 +1,148 @@
...
+
+    try:
+        tbmaster = TimebaseMasterClient(
+            args.client_name,
+            bpm=max(0.0, min(300.0, args.tempo)),
```
mgeier: Why are you silently limiting this instead of raising an error?

Why are you limiting it at all?

I didn't know that, but JACK seems to allow negative BPM values ...

examples/timebase_master.py
```diff
@@ -0,0 +1,148 @@
...
+            beats_per_bar=beats_per_bar,
+            beat_type=beat_type,
+            ticks_per_beat=args.ticks_per_beat,
+            debug=args.debug)
+    except jack.JackError as exc:
+        return "Could not create timebase master JACK client: %s" % exc
```
mgeier: Here you could use `ap.exit(...)`.

https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.exit

examples/timebase_master.py
```diff
@@ -0,0 +1,148 @@
...
+    with tbmaster:
+        if tbmaster.become_timebase_master(args.conditional):
+            try:
+                print("Press Ctrl-C to quit...")
+                tbmaster.stop_event.wait()
+            except KeyboardInterrupt:
+                print('')
+            finally:
+                try:
+                    tbmaster.release_timebase()
```
mgeier: There seems to be an inconsistency:

* `client.set_sync_callback(cb)` vs. `client.set_sync_callback(None)`
* `client.set_timebase_callback(cb)` vs. `client.release_timebase()`

Or am I missing something?

This is of course directly taken from JACK, but should we fix this as long as `release_timebase()` is still very fresh?

But I guess the problem is the second argument, right?

examples/timebase_master.py
```diff
@@ -0,0 +1,148 @@
...
+                except jack.JackError:
+                    # another JACK client might have grabbed timebase master
+                    pass
+        else:
+            return "Timebase master already present. Exiting..."
+
+
+if __name__ == '__main__':
+    sys.exit(main(sys.argv[1:]) or 0)
```
mgeier: Since this is the default behavior, you don't really have to pass anything to `main()`, right?

mgeier:
You have already changed a few double quotes to single quotes, but for my taste you could change even more.

I haven't been completely consistent TBH, but #77 should improve this.

SpotlightKid:
Thanks for the review!

I'm currently a bit busy, but I will try to address these issues / questions in a few days.

examples/timebase_master.py
```diff
@@ -0,0 +1,148 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
```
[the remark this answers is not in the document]
SpotlightKid: Old habit. Fixed in b9ec7aa96879a3284c1a6664a121adf4f064b33d.

examples/timebase_master.py
```diff
@@ -0,0 +1,148 @@
+#!/usr/bin/env python
```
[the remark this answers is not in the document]
SpotlightKid: Fixed in b10725145ca454ad6421a0e5ecb11db0418328f7

examples/timebase_master.py
```diff
@@ -0,0 +1,148 @@
...
+
+import argparse
+import sys
+from threading import Event
+
+import jack
+
+
+class TimebaseMasterClient(jack.Client):
+    def __init__(self, name, bpm=120.0, meter='4/4', beats_per_bar=4,
```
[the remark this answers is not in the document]
SpotlightKid: Done in ead365ded3cc3050166a1f8f80971dc6cc69f4a5

examples/timebase_master.py
SpotlightKid: `ArgumentParser.error()` terminates the script, which is unwanted here for reasons explained below your comment on the last line of the script.

examples/timebase_master.py
SpotlightKid: Because BPM values above 300 aren't very musically useful, IMHO.

I could have implemented a custom `argparse` argument checker, but it seems overkill here. It's just an example, after all.

And how would a negative BPM work? Come to think of it, 0 BPM makes no sense either.

I changed the clamping to the range `0.1 - 300.0` and updated the param help to show the min/max value in da755992a6c251d2503c6e806de7d91f56d5efca.

examples/timebase_master.py
SpotlightKid: Same reason not use it as for `ArgumentParser.error()`

examples/timebase_master.py
SpotlightKid: I don't think we should change names taken from libjack without a good reason (except the removal of the `jack_` prefix). It makes it harder to use the libjack API docs.

We could of course, in addition to having `release_timebase`, augment `set_timebase_callback` to call `release_timebase` if called with a `None` argument, but that would have to be another PR and this script wouldn't need to change.

examples/timebase_master.py
SpotlightKid: I use the following strategy for all my command line scripts (imports not shown):

```py
def main(args=None):
    ap = argparse.ArgumentParser()
    args = ap.parse_args(args if args is not None else sys.argv[1:])

if __name == '__main__':
    sys.exit(main(sys.argv[1:]) or 0)
```

This allows the following:

* You can call the script directly with `python myscript.py` (args are passed via the call to `main()`).
* You can use the script as a `console_script` entry point (`console_scripts = ['myscript = mypkg.myscript:main]`) (`main()` is called without arguments and the command line args have to be retrieved via `sys.argv`).
* You can `import` the script in another Python file and call the `main()` function manually, passing the command line as a string. 
* The `sys.exit(main(...) or 0)` allows the `main()` function to just return an exit code or an error string to set the program exit code or print an error code to `stderr`, but when you call `main()` manually, you can still use the return value (`ArgumentParser.parse_args()` may still terminate the program, so this isn't perfect). Returning `None` (the default with no `return` statement), sets the exit code to 0 (success).

examples/timebase_master.py
```diff
@@ -0,0 +1,148 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#
+#  timebase_master.py
```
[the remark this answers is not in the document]
SpotlightKid: I think it's useful to have this, e.g. when editing in a console editor like `vi`, but if you want, I can remove it.

examples/timebase_master.py
```diff
@@ -0,0 +1,148 @@
...
+    def __init__(self, name, bpm=120.0, meter='4/4', beats_per_bar=4,
+                 beat_type=4, ticks_per_beat=1920, conditional=False,
+                 debug=False, **kw):
+        super().__init__(name, **kw)
+        self.beats_per_bar = beats_per_bar
+        self.beat_type = beat_type
+        self.bpm = bpm
+        self.conditional = conditional
+        self.debug = debug
+        self.ticks_per_beat = int(ticks_per_beat)
```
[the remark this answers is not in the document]
SpotlightKid: That doesn't make much sense to me, musically, but I can convert the instance attribute via `float()` when setting the `pos` attribute.

Also, since tick is an `int32_t`, we need to be careful when assigning expressions to it that involve `ticks_per_beat`, that the type of the expression result is an `int`, otherwise `cffi` will throw an error.

I noticed that `beats_per_bar` and `beat_type` are also of type `float` (again, makes no sense to me, musically), so I changed the code to convert these as well when assigning to `pos`, while keeping the instance attributes as type `int`: e5c018e40d5ad75328f5dc723aefd02a7bbfdac5

examples/timebase_master.py
```diff
@@ -0,0 +1,148 @@
...
+        self.stop_event = Event()
+        self.set_shutdown_callback(self.shutdown)
+
+    def shutdown(self, status, reason):
```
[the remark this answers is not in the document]
SpotlightKid: Since `TimebaseMasterClient.stop_event` is public, I figured `shutdown` should be too. Agreed, encapsulation could be better, e.g. by making `stop_event` private and adding a method (e.g. `run()`) that does the whole business of reseting the event, setting the shutdown and timebase callbacks and then waiting for the event.

examples/timebase_master.py
```diff
@@ -0,0 +1,148 @@
...
+            print("New pos:", jack.position2dict(pos))
+
+        # Adapted from:
+        # https://github.com/jackaudio/jack2/blob/develop/example-clients/transport.c#L66
+        if new_pos:
+            pos.beats_per_bar = self.beats_per_bar
+            pos.beats_per_minute = self.bpm
+            pos.beat_type = self.beat_type
+            pos.ticks_per_beat = self.ticks_per_beat
+            pos.valid |= jack._lib.JackPositionBBT
```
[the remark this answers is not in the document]
SpotlightKid: Just defensive programming. I wasn't sure wether other fields of `pos` (e.g timecode fields) could be set from somewhere else, so I'm just updating the `JackPositionBBT' bit, leaving any other bits intact.

examples/timebase_master.py
```diff
@@ -0,0 +1,148 @@
...
+        self.stop_event = Event()
+        self.set_shutdown_callback(self.shutdown)
+
+    def shutdown(self, status, reason):
+        print('JACK shutdown:', reason, status)
+        self.stop_event.set()
+
+    def _tb_callback(self, state, nframes, pos, new_pos):
+        if self.debug and new_pos:
+            print("New pos:", jack.position2dict(pos))
```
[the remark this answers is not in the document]
SpotlightKid: Since the program changes `pos` itself, I think it makes more sense for debugging to print what gets passed.

examples/timebase_master.py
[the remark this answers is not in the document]
SpotlightKid: `dict2position` could be handy, yes, but I think that should be in a separate PR.

examples/timebase_master.py
```diff
@@ -0,0 +1,148 @@
...
+        if new_pos:
+            pos.beats_per_bar = self.beats_per_bar
+            pos.beats_per_minute = self.bpm
+            pos.beat_type = self.beat_type
+            pos.ticks_per_beat = self.ticks_per_beat
+            pos.valid |= jack._lib.JackPositionBBT
+
+            minutes = pos.frame / (pos.frame_rate * 60.0)
+            abs_tick = minutes * pos.beats_per_minute * pos.ticks_per_beat
+            abs_beat = abs_tick / self.ticks_per_beat
```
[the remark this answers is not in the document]
SpotlightKid: Truncation to int happens below. I don't think it makes a difference.

examples/timebase_master.py
```diff
@@ -0,0 +1,148 @@
...
+
+            pos.bar = int(abs_beat / pos.beats_per_bar)
+            pos.beat = int(abs_beat - (pos.bar * pos.beats_per_bar) + 1)
+            pos.tick = int(abs_tick - (abs_beat * pos.ticks_per_beat))
+            pos.bar_start_tick = pos.bar * pos.beats_per_bar * pos.ticks_per_beat
+            pos.bar += 1  # adjust start to bar 1
+        else:
+            # Compute BBT info based on previous period.
```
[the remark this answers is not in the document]
SpotlightKid: My interpretation of http://jackaudio.org/api/structjack__position__t.html is that `pos.valid` is only set by the timebase master. I'll ask around whether that's true.

examples/timebase_master.py
```diff
@@ -0,0 +1,148 @@
...
+
+                if pos.beat > pos.beats_per_bar:
+                    pos.beat = 1
+                    pos.bar += 1
+                    pos.bar_start_tick += pos.beats_per_bar * pos.ticks_per_beat
+
+                    if self.debug:
+                        print("Pos:", jack.position2dict(pos))
+
+    def become_timebase_master(self, conditional=None):
```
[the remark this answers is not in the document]
SpotlightKid: Instance default and per-call value. If using object-orientation, why not making good use of it?

examples/timebase_master.py
SpotlightKid: Old habit. I don't think `argparse` always did that. But in Python 3 it should always work. Fixed in 60e57a89d78e329b247a80788f70ae8271c97db2.

examples/timebase_master.py
SpotlightKid: You get a better error, because `split` always yields two values. If there is an extra `/` in `args.meter`, the second value will include it and `int(x)` will fail.

examples/timebase_master.py
SpotlightKid: Old habit and it's often shorter and quicker to type. Fixed in 29603b1560fced89e80738b3202b76e0863ef336.

examples/timebase_master.py
SpotlightKid: As discussed above `ap.parse_args(args if args is not None else sys.argv[1:])` can be shortened to `ap.parse_args(args)`.

examples/timebase_master.py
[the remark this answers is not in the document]
SpotlightKid: Fixed in f03487e8e7d5d8f5a61d2b59d9fffb38396b9175.

SpotlightKid:
@mgeier Please review again.

examples/timebase_master.py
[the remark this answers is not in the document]
mgeier: I don't think this is a good enough reason.

You can use `:f` to display the file name in `vi`.

Then you don't even have to scroll up to find out the file name.

examples/timebase_master.py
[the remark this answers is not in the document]
mgeier: > That doesn't make much sense to me, musically

Probably not, probably it does?

I would simply allow `float` values because JACK allows them.

> we need to be careful when assigning expressions to it

Yes, but it doesn't help if we assign an `int` to `ticks_per_beat`, because it will be immediately converted to a `double`.

> again, makes no sense to me, musically

Again, I don't know, I would still allow it if JACK thinks that's a good idea.

> keeping the instance attributes as type `int`

I would just pass it through to the JACK functions, CFFI will check it and complain if it isn't right.

I think converting the inputs with `int()` is not a good idea, because it silently converts user-provided `float` values.
If you really want to require `int`s (which I think you shouldn't), you should at least fail loudly when the uses tries to use a `float` value.

Also, it's very confusing if your local copy of the values have different data types than the values in the `pos` struct, isn't it?

examples/timebase_master.py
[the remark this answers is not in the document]
mgeier: The more I think about it, the less I understand ...

I have the feeling that the JACK API isn't 100% consistent in itself and this isn't possible to get right without some additional assumptions about the behavior of other clients.

I would keep it the same as the JACK example, but if you are confident `|=` is the right thing to do, that's also fine by me.

> I think that should be in a separate PR.

Yes, sure.

But that PR would have to be merged before this PR?

examples/timebase_master.py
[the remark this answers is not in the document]
mgeier: OK

examples/timebase_master.py
mgeier: > You get a better error

    ValueError: invalid literal for int() with base 10

vs.

    ValueError: too many values to unpack (expected 2)

I would argue that the second is better, especially since the error for too few `/` characters is very similar in both situations.

But it doesn't matter at all, since you are catching this exception, and nobody ever sees it, right?

examples/timebase_master.py
```diff
@@ -0,0 +1,146 @@
...
+        'tempo',
+        nargs='?',
+        type=float,
+        default=120.0,
+        help="Tempo in beats per minute (0.1-300.0, default: %(default)s)")
+
+    args = ap.parse_args(args)
+
+    try:
+        beats_per_bar, beat_type = (int(x) for x in args.meter.split('/', 1))
```
mgeier: BTW, this would be a perfect opportunity to use `map()`:

```python
map(int, args.meter.split('/'))
```

examples/timebase_master.py
mgeier: > aren't very musically useful

maybe, maybe not.

I think it is much more interesting to see how JACK reacts to certain values, than what restrictions were added in this example.

> I could have implemented a custom `argparse` argument checker

I wouldn't check the range at all.

> And how would a negative BPM work? Come to think of it, 0 BPM makes no sense either.

I don't know any of this.

But I would certainly be interested how JACK handles those values, and if you are limiting them in the example, I can't easily find out.

examples/timebase_master.py
mgeier: I think there are two different things:

* calling `main()` vs `main(sys.argv[1:])` (your first 3 points)
* terminating on error or returning something (your last point)

AFAICT, there is no disadvantage in calling `main()` without arguments here, it works exactly the same in all mentioned cases.

Your last point is a separate thing.

I guess it's possible to do it that way, but I'm not sure if you are gaining anything. It probably even makes things worse.

On one hand, it might be a YAGNI situation just adding useless noise.
Did you ever hear of a usage where this mattered?

On the other hand, I'm not sure what you would even do when you get an "error value" except for terminating the program.
Is there a conceivable situation where you wouldn't want to terminate the program?

And even if there is a very teeny tiny chance that somebody reeeeally doesn't want to terminate, they could still catch `SystemExit`, right?

> (`ArgumentParser.parse_args()` may still terminate the program, so this isn't perfect)

Exactly, far from it.

It's actually an extremely common case.

And I think this is the main argument here.
Why should a wrong `--tempo` argument take a completely different code path than a wrong `--meter` argument?

All in all, it sounds like much noise for little gain, doesn't it?

examples/timebase_master.py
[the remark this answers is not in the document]
mgeier: OK, sounds interesting.

What about slapping an `assert` in there for now, just to see if it ever happens?

We can remove it if it turns out to be wrong.

examples/timebase_master.py
mgeier: ... or of course with `float`, if we choose to allow that.

mgeier:
Sorry @SpotlightKid, I kinda forgot about this PR ...

Thanks for your recent changes, what do you think about my still open comments?

There is still a lack of an official way to use `jack._lib.JackPositionBBT`. Do you want to implement this or should I?
Do you have a suggestion how we should do this?

SpotlightKid:
Since `jack_position_bits_t` is a bit field, i.e. values are meant to be ORed together, I would just export these as module level constants:

```
JACK_POSITION_BBT = _lib.JacKPositionBBT
JACK_POSITION_TIMECODE = _lib.JackPositionTimecode
JACK_POSITION_BBT_FRAME_OFFSET _lib.JackBBTFrameOffset
JACK_POSITION_AUDIO_VIDEO_RATIO = _lib.JackAudioVideoRatio
JACK_POSITION_VIDEO_FRAME_OFFSET = _lib.JackVideoFrameOffset
```

SpotlightKid:
> what do you think about my still open comments?

TBH, I don't really care too much, just change these things it in the way you like.

examples/timebase_master.py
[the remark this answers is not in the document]
mgeier: I've created a possible implementation of `dict2position()` in #78.

mgeier:
> > what do you think about my still open comments?
>
> TBH, I don't really care too much, just change these things it in the way you like.

OK, cool, will do.

But I'd first like to hear your opinion on which of those options is the best:

* add `dict2position()`, see #78 
* add `JACK_POSITION_*` module constants, as mentione above (https://github.com/spatialaudio/jackclient-python/pull/74#issuecomment-549089422)
* add both

I have the feeling that adding only one of them would suffice for now.
If needed, the other one could be added later.

SpotlightKid:
I personally would prefer `JACK_POSITION_*` constants. I find dict-based interfaces cumbersome and ugly to use, and there's the danger of mistyping dict keys.

mgeier:
I was also skeptical about #78 and created a branch with the `JACK_POSITION_*`. I just seem to have forgotten to make PR, which I've done just now: #82.

mgeier:
@SpotlightKid I've re-based and squashed your commits from this PR into #83 and added a few commits there.

PR #91, opened 2020-08-12
Use jack_port_rename instead of deprecated jack_port_set_name

lack:


mgeier:
Thanks for this PR!

You can ignore the first case mentioned by the PEP8 bot, but can you please fix the other two?

jack_build.py
```diff
@@ -141,7 +141,8 @@
 const char** jack_port_get_all_connections(const jack_client_t* client, const jack_port_t* port);
 /* deprecated: jack_port_tie */
 /* deprecated: jack_port_untie */
-int jack_port_set_name(jack_port_t* port, const char* port_name);
```
mgeier, line 144: You could leave the declaration in there (but also keep the comment marking it as deprecated).

This way, users who would like to use it (for whatever reason) can still call it from their code using the `_lib` object.

And I think the declaration doesn't hurt, even if in the future this function will completely be removed from the JACK library.

src/jack.py
```diff
@@ -1979,8 +1980,7 @@ class OwnPort(Port):
     """
 
     def __init__(self, port_ptr, client):
-        Port.__init__(self, port_ptr)
-        self._client = client
+        Port.__init__(self, port_ptr, client)
```
mgeier, line 1983: I think you can just remove the whole `__init__()` method from `OwnPort`, right?

src/jack.py
```diff
@@ -1783,9 +1783,9 @@ def _wrap_port_ptr(self, ptr):
         """Create appropriate port object for a given port pointer."""
         porttype = _ffi.string(_lib.jack_port_type(ptr))
         if porttype == _AUDIO:
-            port = OwnPort(ptr, self) if self.owns(ptr) else Port(ptr)
+            port = OwnPort(ptr, self) if self.owns(ptr) else Port(ptr, self)
         elif porttype == _MIDI:
-            port = OwnMidiPort(ptr, self) if self.owns(ptr) else MidiPort(ptr)
+            port = OwnMidiPort(ptr, self) if self.owns(ptr) else MidiPort(ptr, self)
```
mgeier, line 1788: Since all classes now have the same initialization arguments, you could refactor this thing:

```python
        if porttype == _AUDIO:
            cls = OwnPort if self.owns(ptr) else Port
        elif porttype == _MIDI:
            cls = OwnMidiPort if self.owns(ptr) else MidiPort
        else:
            assert False
        return cls(ptr, self)
```

What do you think?

HaHeho: Also, `assert` statements should only be used in testing not in production code (according to some recommendation, I would have to look it up if you want to know).

Also, they get ignored when executing with the Python command line option `-o` (request optimization).

I suppose both points are arguments that the `assert` should be refactored as well. However, there is (probably) no practical benefit to it right now. Nevertheless, this would not be the PR to do so, since there is other code sections where `assert` is used as well.

mgeier: @HaHeho I agree with you general statements, but I disagree that this applies here.

IMHO using `assert` here is exactly the right thing to do.

My rule for `assert` is: users should never see an assertion error if they only use the public API. If they see an assertion error nevertheless, there is a bug in the library, it's never the user's fault (except if they use underscore-prefixed "hidden" stuff or monkey-patching or something like that).

Now in this case, the `porttype` comes from the JACK library, and our code assumes that there are only two possible port types. I'm quite sure this is true right now, but it might theoretically change in the future. If and when that happens, our code will have a bug. But luckily, the `assert` statement will immediately inform us about it.
Probably even more importantly, it informs a human reader of the source code that we didn't just forget the `else` clause.

You didn't suggest an alternative, I could think of those two:

* `raise` a "proper" exception with a nice message for the user. Since the user *should* never see this, this is kinda superfluous, isn't it? *If* a user sees this nevertheless, they did nothing wrong, there is nothing they can do. They should just report it to the maintainers, like every other error they don't understand.

* Remove the `else` clause. This doesn't change anything if everything goes well. And if there is a bug, it just makes the bug less obvious. And for a reader of the code, it makes the programmer intention less obvious (did they deliberately leave out the `else` clause or did they just forget it?).

Do you know any other alternatives?

> Also, they get ignored when executing with the Python command line option `-o` (request optimization).

And that's not a big deal.
If there is no bug, nothing changes (except it's faster!).
If there *is* a bug, the error message will just be less clear. In some extreme cases (but not here), this might lead to no error being raised and a wrong result to be returned. But it's a buggy program, so no surprise there. That's the whole reason why there is an `-o` option in the first place. If it wouldn't make a difference, it would be enabled unconditionally all the time.

> there is other code sections where `assert` is used as well.

And AFAICT, all of them are used deliberately in situations like I just described.

If not, please let me know (e.g. by making a PR)!

HaHeho: The only true alternative is replace it with an Exception, which is also more literate.

However, as you also said, the code is not supposed to be executed under regular conditions anyways. So if you don't want it to be deliberately avoided, no work is required.

lack:
Will do!

lack:
My latest changes address all the suggestions except the one about the "assert" which predates my change here.

jack_build.py
HaHeho: @lack the deprecated info for the old function would still be desirable. This is still missing.

`int jack_port_set_name(jack_port_t* port, const char* port_name);  /* deprecated */`

mgeier:
@HaHeho 

> Seems nice now. Except for maybe the @pep8speaks linter warnings.

Those lines are in the C declarations, I don't care about long lines there.

> Are they updated automatically with new commits, or how does it work?

Yes.

It's a great service, I can recommend it to any Python project: https://pep8speaks.com/

The comment above (https://github.com/spatialaudio/jackclient-python/pull/91#issuecomment-672450759) tells you the last change: "Comment last updated at 2020-08-14 21:59:43 UTC".
You can even look at the edit history of that comment to see all changes.

PR #162, opened 2026-09-22
Restore optional NumPy dependency

adamtheturtle:
The move from `setup.py` to `pyproject.toml` in 0.5.6 dropped the `NumPy` extra. Restore it for callers of `OwnPort.get_array()`, which imports NumPy at runtime.

This also restores the published dependency metadata that `types-JACK-Client` relies on: its `get_array()` stub returns `numpy.typing.NDArray`, so the stubs package depends on NumPy. The missing extra is currently failing typeshed's stub uploader tests on `main` and on unrelated PRs. A new PyPI release after this change is needed to unblock those checks.

Verified with `uv build --wheel`: the wheel metadata now contains `Provides-Extra: numpy` and `Requires-Dist: numpy; extra == "numpy"`.


JelleZijlstra:
This also needs a uv.lock update.

Also, note that none of the JACK-Client documentation recommends using this `numpy` extra. Having the extra makes our lives a bit easier on typeshed, but it's not necessary.

mgeier:
Thanks for this PR!

Can you update uv.lock or should I do it?

mgeier:
> A new PyPI release after this change is needed to unblock those checks.

Done: https://pypi.org/project/JACK-Client/0.5.7/

